PR #7, opened 2018-03-14
BASW-65: Add In Arrears Events to Membership Status Rules

MiyaNoctem:
## Overview
To manage membership status for those payed with a payment plan in several instalments, we need two events that will allow to check for when programmed payments are overdue:

- In arrears: triggered when a membership is paid by recurring contribution with instalments and at least one pending instalment contribution's receive date falls to the past.
- Not in arrears: no pending instalment contribution in the past.

## Before
There was no way to manage the status of a membership payed with instalments in regard to the status of those payments.

## After
The two new events (In Arrears, Not In Arrears) were added as options to membership status form edit view, on **Start Event** and **End Event** fields.

Hook alterCalculatedMembershipStatus was implemented to go through the list of status rules sorted by weight, check if any of them use arrears related events as start or end, and validating for those that do if start and end conditions are met to set the status.

### DIFF
--- a/CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
+++ b/CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
@@ -35,9 +35,6 @@ class CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus {
    */
   private static $contributionStatusValueMap = [];
 
-  /**
-   * CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus constructor.
-   */
   public function __construct() {
     if (count(self::$memberShipStatuses) == 0) {
       $membershipStatuses = civicrm_api3('MembershipStatus', 'get', [
--- a/CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
+++ b/CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
@@ -6,11 +6,38 @@
  */
 class CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus {
 
+  /**
+   * Array with the membership's details, as passed by the hook function.
+   *
+   * @var array
+   */
   private $membership;
-  private $calculationArguments;
-  private static $memberShipStatuses = array();
-  private static $contributionStatusValueMap = array();
 
+  /**
+   * Dates used by CiviCRM to calculate status, ie. start_date, end_date and
+   * join_date.
+   *
+   * @var array
+   */
+  private $calculationArguments;
+
+  /**
+   * List of all available membership statuses.
+   *
+   * @var array
+   */
+  private static $memberShipStatuses = [];
+
+  /**
+   * Maps the names of contribution statuses to their corresponding values.
+   *
+   * @var array
+   */
+  private static $contributionStatusValueMap = [];
+
+  /**
+   * CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus constructor.
+   */
   public function __construct() {
     if (count(self::$memberShipStatuses) == 0) {
       $membershipStatuses = civicrm_api3('MembershipStatus', 'get', [
@@ -21,9 +48,9 @@ class CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus {
     }
 
     if (count(self::$contributionStatusValueMap) == 0) {
-      $contributionStatuses = civicrm_api3('OptionValue', 'get', array(
+      $contributionStatuses = civicrm_api3('OptionValue', 'get', [
         'option_group_id' => "contribution_status",
-      ));
+      ]);
 
       foreach ($contributionStatuses['values'] as $currentStatus) {
         self::$contributionStatusValueMap[$currentStatus['name']] = $currentStatus['value'];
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+   *   Date to use as reference to check start date
+   * @param string $adjustUnit
+   *   Unit of time interval to be added to start date
+   * @param int $adjustInterval
+   *   Amount of time to be added to start date
+   *
+   * @return boolean
+   *   True if the event has occurred, false otherwise
+   */
+  private function checkEvent($event, $referenceDate, $adjustUnit, $adjustInterval) {
```
omarabuhussein: due to this check https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/824cd2cf95b42edeb2a3c4ac7f19f73035d4c641/CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php#L52

the status to be passed to checkEvent method can be either in_arrears or not_arrears, so I don't see a reason to add a case to other types, and based on that we can remove this method and call checkInArrearsEvent and checkNotInArrearsEvent methods directly.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+        // If we reach calculated status, we don't need to consider other options by weight.
+        if ($calculatedStatus['id'] == $currentStatus['id']) {
+          break;
+        }
+
+        // No arrears, so we continue with next status.
+        continue;
+      }
+
+      $startEvent = $this->checkEvent(
```
omarabuhussein: so as I explained in my other comment, you can just call checkInArrearsEvent here directly.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+        CRM_Utils_Array::value('start_event', $currentStatus),
+        $arguments['status_date'],
+        CRM_Utils_Array::value('start_event_adjust_unit', $currentStatus),
+        CRM_Utils_Array::value('start_event_adjust_interval', $currentStatus)
+      );
+
+      $endEvent = $this->checkEvent(
```
omarabuhussein: and here u can just call checkNotInArrearsEvent directly

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+        CRM_Utils_Array::value('end_event', $currentStatus),
+        $arguments['status_date'],
+        CRM_Utils_Array::value('start_event_adjust_unit', $currentStatus),
+        CRM_Utils_Array::value('start_event_adjust_interval', $currentStatus)
+      );
+
+      if ($startEvent && !$endEvent) {
```
omarabuhussein: I don't see the point of !$endEvent here to be honest, at the end it is just the inverse of $startEvent check so I don't think we need it all.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+ */
+class CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus {
+
+  private $membership;
+  private $calculationArguments;
+  private static $memberShipStatuses = array();
+  private static $contributionStatusValueMap = array();
+
+  public function __construct() {
+    if (count(self::$memberShipStatuses) == 0) {
```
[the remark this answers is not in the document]
MiyaNoctem: The alterCalculatedMembershipStatus hook is called from the job to update all memberships statuses, so basically, it's called once for every active membership in the system.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+   *   Arguments passed in to calculate date
+   * @param array $membership
+   *   Membership details from the calling function
+   */
+  public function alterMembershipStatus(&$calculatedStatus, $arguments, $membership) {
+    $this->membership = $membership;
+    $this->calculationArguments = $arguments;
+
+    foreach (self::$memberShipStatuses as $currentStatus) {
+      $startEventIsArrearsRelated = stripos($currentStatus['start_event'], 'arrears') !== false;
```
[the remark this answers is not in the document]
MiyaNoctem: Yeah, I wanted to capture both 'In arrears' and 'Not in arrears' in a single instruction, and not worry about upper and lower case letters.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -0,0 +1,336 @@
...
+      $endEventIsArrearsRelated = stripos($currentStatus['end_event'], 'arrears') !== false;
+
+      if (!$startEventIsArrearsRelated && !$endEventIsArrearsRelated) {
+        // If we reach calculated status, we don't need to consider other options by weight.
+        if ($calculatedStatus['id'] == $currentStatus['id']) {
```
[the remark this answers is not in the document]
MiyaNoctem: Statuses are verified in order by their weight, and once a status that meets start and end criteria is found, the rest are ignored. The only reason to check a status in this hook is if it uses arrears related events for either start or end, in case it was falsely determined it did not meet start and end criteria.

Since in this case: 
- I have already processed all past statuses and found that none of them meet start and end events
- The current status does not use 'In arrears' or 'Not in arrears' events
- CiviCRM has already evaluated that this status meets both start and end criteria

I can safely deduce this is the actual status of the membership and there is no need to continue processing the rest of possible statuses, as they should only be evaluated if all past statuses do not meet start and end criteria.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
MiyaNoctem: I checked with Guanhuan and he said it was ok to mix arrears events with other events, so for example, I could create a status called 'Urgent Collection', starting one month before 'End date' and finishing in 'Not in arrears', so memberships with less than one month to go, with outstanding invoices for a payment plan are marked.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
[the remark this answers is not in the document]
omarabuhussein: but civicrm API return the machine name for the start and end event, so in this case it can be either in_arrears or not_arrears since these are the machine names u used, correct ? 
 
If I am not missing anything, it will be more robust to use the exact machine name rather than stripos.

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -6,11 +6,38 @@
...
+
+  /**
+   * Maps the names of contribution statuses to their corresponding values.
+   *
+   * @var array
+   */
+  private static $contributionStatusValueMap = [];
+
+  /**
+   * CRM_MembershipExtras_Hook_Alter_CalculatedMembershipStatus constructor.
```
omarabuhussein: no need for this docblock

CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
MiyaNoctem: Needed for combination of arrears related events with other events.

PR #8, opened 2018-03-22
BASW-68: Adding scheduled job to renew offline auto renewal memberships

omarabuhussein:
A scheduled job  called“Renew offline auto-renewal memberships” is added.

When the scheduled job run, it finds and processes any membership that meets all following conditions:

1. Its last recurring contribution is using offline payment processor
2. Its last recurring contribution is set to auto-renew
3. Its last recurring contribution is not cancelled or refunded.
4. The membership's end date is equals to or smaller than today.

- If the membership's last recurring contribution has no installments (installments = NULL) then it will  generate a new pending contribution under that recurring contribution. Then it will extend the membership for one term.

- If the membership's last recurring contribution has installments (installments != NULL), the recurring contribution will be used as a template to create a new recurring contribution with current date and pending contributions = number of recurring contribution installments linking to the new  recurring contributions.  Then extend the membership for one term.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal.php
@@ -1,0 +1,385 @@
+<?php
+
+class CRM_MembershipExtras_Job_OfflineAutoRenewal {
+
+  /**
+   * Mapping between financial types IDs and Machine Names.
+   *
+   * @var array
+   */
+  private $financialTypesIDMap = [];
+
+
+  /**
+   * The ID of the membership that currently
+   * being renewed/processed.
+   *
+   * @var int
+   */
+  private $currentMembershipID;
+
+  /**
+   * The ID of the recurring Contribution linked
+   * with the membership that currently
+   * being renewed/processed.
+   *
+   * @var int
+   */
+  private $currentRecurContributionID;
+
+  /**
+   * The data of the last Contribution related to
+   * the membership to-be-renewed recurring
+   * contribution.
+   *
+   * We use it as a base for creating the
+   * new contribution for the auto-renewed
+   * membership.
+   *
+   * @var int
+   */
+  private $lastContribution = [];
+
+  /**
+   * The ID of contact who his
+   * membership is to be auto-renewed.
+   *
+   * @var int
+   */
+  private $contactID;
+
+  public function __construct() {
+    $this->setFinancialTypesIDMap();
+  }
+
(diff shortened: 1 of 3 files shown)


jamienovick:
@KarinG @eileenmcnaughton Just cc the above so you know that this is another part of what we are developing. Obviously we are doing this outside core based on earlier feedback but in case you think it relevant to our other conversation

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -0,0 +1,385 @@
...
+      'contribution_id' => $this->lastContribution['id'],
+      'options' => ['limit' => 1, 'sort' => 'id DESC'],
+    ])['values'][0];
+
+    $lineItemID = civicrm_api3('LineItem', 'create', [
+      'sequential' => 1,
+      'entity_table' => 'civicrm_membership',
+      'entity_id' => $this->currentMembershipID,
+      'contribution_id' => $contributionID,
+      //'label' => '' TODO : Student (33.33%)March 19th, 2018,
```
MiyaNoctem: You still need to do this right?

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -0,0 +1,385 @@
...
+      'sequential' => 1,
+      'return' => ['id', 'amount', 'currency'],
+      'entity_id' => $lastContributionLineItem['id'],
+      'entity_table' => 'civicrm_line_item',
+      'options' => ['limit' => 1, 'sort' => 'id DESC'],
+    ])['values'][0];
+
+    civicrm_api3('FinancialItem', 'create', [
+      'contact_id' => $this->contactID,
+      //'description' => '' TODO : Student (33.33%)March 19th, 2018,
```
MiyaNoctem: Again, this should be resolved before merging, right?

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
omarabuhussein: done

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
omarabuhussein: done

omarabuhussein:
I will merge this for now and do some cleaning work in BASW-90

PR #9, opened 2018-03-23
BASW-67: Implement Payment Plan on Membership Create and Renewal

MiyaNoctem:
## Overview
In the payment section of "New Membership" and "Renew Membership" modals, two options should be provided: 
- Contribution: Contribution option is the traditional way of paying for a membership which is by creating a single contribution. This option keeps the original behaviour of how it works currently in CiviCRM.
- Payment Plan: Payment Plan option allows user to pay for the membership in instalments.

With Payment Plan option, the following fields should be provided:
- Pay by a different contact? - allow contact other than the beneficiary to pay and also soft credit support
- Amount - the price of the selected membership.
- Financial Type - the financial type the recurring contribution and contribution should use
- Number of Instalments - number of instalments the full price should be slipt into
- Interval and interval unit - the gap between two instalments
- Payment Method - payment method used for the instalment contibutions
- Payment Plan Start Date - Date of the first instalment contribution

## Before
Memberships were only able to be paid for in full.

## After
It is now possible to choose to create a payment plan to pay for a membership, both on membership creation and renewal forms.

### Altered new membership and renewal forms
Added a radio button to choose if membership is to be paid for in full, or using a payment plan. Choosing payment plan, disables some fields and adds others that are needed to configure the payment plan.

#### New Membership
bas-67-new-membership-2

#### Membership Renewal
bas-67-renew-membership-2

### Payment Plan creation
Once a membership is created/renewed, post-process hooks were implemented to mutate the membership and the contribution into one payed for by a payment plan by:

- Creating recurring contribution
- Creating a new contribution for each installment, for the corresponding percentage of the full amount
- Deleting the original contribution for the full amount


CRM/MembershipExtras/Hook/PostProcess/MembershipRenewal.php
```diff
@@ -0,0 +1,35 @@
...
+  public function postProcess() {
+    $isMembershipRenewal = $this->form->getAction() & CRM_Core_Action::RENEW;
+    $recordingContribution = $this->form->getSubmitValue('record_contribution');
+    $contributionIsPaymentPlan = $this->form->getSubmitValue('contribution_type_toggle') == 'payment_plan';
+
+    if ($isMembershipRenewal && $recordingContribution && $contributionIsPaymentPlan) {
+      $this->loadCurrentMembershipAndContribution();
+      $this->createRecurringContribution();
+      $this->createInstallmentContributions();
+      $this->deleteOldContribution();
```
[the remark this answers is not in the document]
MiyaNoctem: The BAO handles the deletion of the other related stuff, like the activity, financial transactions, line items, etc, which is why I called the API. However you just made me realize the API checks for some specific permissions that a user could not have, so I the implementation of deleteOldContribution() to use the BAO instead.

CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
```diff
@@ -0,0 +1,72 @@
...
+
+  /**
+   * This method should check if the given action is the proper operation so
+   * that the form is changed to have payment plan data (ie. new for membership
+   * form, renew for membership renewal form).
+   *
+   * @param int $action
+   *   Constant value for the action, as defined in CRM_Core_Action class
+   *
+   * @return mixed
```
omarabuhussein: this should be Boolean not mixed

CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
```diff
@@ -0,0 +1,72 @@
...
+  public function buildForm() {
+    $this->addPaymentPlanSection();
+  }
+
+  /**
+   * Adds new fields to the form and includes template that shows the fields and
+   * juggles around exiting ones.
+   */
+  protected function addPaymentPlanSection() {
+    if ($this->isCorrectOperation($this->form->_action)) {
```
omarabuhussein: use guadrd statment instead : 

if (!$this->isCorrectOperation($this->form->_action)) {
return;
}

CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
```diff
@@ -0,0 +1,72 @@
...
+   * This method should check if the given action is the proper operation so
+   * that the form is changed to have payment plan data (ie. new for membership
+   * form, renew for membership renewal form).
+   *
+   * @param int $action
+   *   Constant value for the action, as defined in CRM_Core_Action class
+   *
+   * @return mixed
+   */
+  protected abstract function isCorrectOperation($action);
```
omarabuhussein: I think isAllowedOperation is more clear name than "Correct"

guanhuan:
Hi @MiyaNoctem ,

This looks good in general. Three small points (one of them is a change):

1. Optimise the bottom spacing for "First instalment summary": I can see there is a duplicated section heading "Contribution and Payment Plan" for the "Check Number" field. Can we remove the duplicate section heading and have the check number field looking like a part of the "First instalment summary" section?
2. For the recurring contribution that is generated after save this with payment plan option, can we change to use the instalment amount rather than the total amount as per A/C stated? That is: to use £10 instead of £120. There has been a change to our general approach, we will not be multiplying the value for recurring contribution amount any more.
3. Can the Gifs include the end results that are produced?

CRM/MembershipExtras/Hook/PostProcess/MembershipPaymentPlan.php
```diff
@@ -0,0 +1,418 @@
...
+  /**
+   * This method should check if the given action is the proper operation so
+   * that the form is post-processed.
+   *
+   * @param int $action
+   *   Constant value for the action, as defined in CRM_Core_Action class
+   *
+   * @return mixed
+   */
+  protected abstract function isCorrectOperation($action);
```
omarabuhussein: same here, I think isAllowedOperation is a more clear name

CRM/MembershipExtras/Hook/PostProcess/MembershipRenewal.php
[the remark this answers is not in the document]
omarabuhussein: but I think calling the API from php does not do permission check, right ?

omarabuhussein:
@MiyaNoctem  I left some small notes

CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
MiyaNoctem: Allowed is not the adequate concept. For example, a user may be allowed to create new memberships, but that's not what I need to verify. What I need to check is if the current action is the action that would require the change, as the same form can be used to 'view', 'update' or 'delete'. I need to make sure action is either 'new' or 'renew', respectively. 

I had to implement this, because the code was getting executed when I tried to delete  membership, for example, as the new membership form is used both to show the delete modal, and do the actual deletion. Crazy, I know, but there's CiviCRM for you :)

So I don't need to check if the action is allowed, but I do need to check if the current action is the correct/adequate/proper one to change the form structure.

CRM/MembershipExtras/Hook/PostProcess/MembershipPaymentPlan.php
MiyaNoctem: See response to prior comment ;)

CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
omarabuhussein: I see, I think it make sense

guanhuan:
I acknowledge Camilo's comment regarding how check number field is handled currently and we will get back to that at some later point. This is good to go for me too.

PR #11, opened 2018-04-10
BASW-81: Add option for using last membership price when renewing offline memberships

omarabuhussein:
When the "**offline memberships auto-renewal scheduled job**" runs, it will use the same price used in the last  payment, but sometimes and for some organizations, the membership price change and using the old price is no longer applicable.

In this PR we add a new configuration page called "**Payment Plan Configuration**" which can be accessed from the administer menu (or **civicrm/admin/payment_plan_settings**) that allow the user to toggle a checkbox called "**Use latest price when auto renew membership?**".

If the checkbox mentioned above is enabled, then the scheduled job will use the membership latest price when doing an auto-renew, if it is disabled then it will use the old price.

aaatest

### DIFF
--- a/CRM/MembershipExtras/Form/PaymentPlanSettings.php
+++ b/CRM/MembershipExtras/Form/PaymentPlanSettings.php
@@ -1,0 +1,108 @@
+<?php
+
+/**
+ * Payment Plan Settings form controller
+ *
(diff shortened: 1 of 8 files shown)


omarabuhussein:
since I am blocked with this  I will have to merge it along with BASW-82 PR : https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/12

PR #15, opened 2018-04-25
BASW-90: Fixing pricesets payments and other fixes

omarabuhussein:
When paying for membership using pricesets, the line items for the contributions were not getting calculated correctly, this is fixes in this PR by using the previous contribution line items as a base for creating the new one.

Also another issue is fixed that prevent you from paying for the a membership using payment plan if you don't open the form from membership contact tab.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal.php
@@ -12,15 +12,6 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal {
    */
   private $financialTypesIDMap = [];
 
-
-  /**
-   * The ID of the membership that currently
-   * being renewed/processed.
-   *
-   * @var int
-   */
-  private $currentMembershipID;
-
   /**
    * The ID of the recurring Contribution linked
    * with the membership that currently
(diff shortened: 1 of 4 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -103,119 +140,76 @@ private function setUseMembershipLatestPrice() {
...
-     $this->calculateTotalAmount($membership);
-     $this->calculateTaxAmount($membership['financial_type_id']);
+     if (empty($recurContribution['installments'])) {
+       $this->currentInstallmentsNumber = 1;
+     }
+     else {
+       $this->currentInstallmentsNumber = $recurContribution['installments'];
+     }
+
+
```
MiyaNoctem: Please remove extra blank line.

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -240,49 +234,192 @@ private function getContributionStatusesNameMap() {
...
+  /**
+   * Renews the payment plan and the related memberships if
    * it paid by installments.
    */
-  private function renewWithInstallmentsMembership() {
-    $previousRecurContributionId = $this->currentRecurContributionID;
+  private function renewWithInstallmentsPaymentPlan() {
     $this->createRecurringContribution();
 
+    $this->recordPaymentPlanFirstContribution();
```
MiyaNoctem: Nice approach, I hated how before we needed to instantiate `MembershipInstallmentsHandler` constructor with two different sets of input parameters to basically do the same thing. Hadn't thought of moving the logic for creating the first installment out of the class. Cool.

membershipextras.php
```diff
@@ -166,22 +166,48 @@ function membershipextras_civicrm_pre($op, $objectName, $id, &$params) {
...
-  $context = CRM_Utils_Request::retrieve('context', 'String');
+  $isPaymentPlanPayment = _membershipextras_isPaymentPlanPayment();
 
-  $membershipContributionCreation = $objectName === 'Contribution' && $op === 'create'
-    && $context === 'membership';
-  $newMembershipContribution = $membershipContributionCreation && ($action & CRM_Core_Action::ADD);
-  $renewMembershipContribution = $membershipContributionCreation && ($action & CRM_Core_Action::RENEW);
+  $membershipContributionCreation = ($objectName === 'Contribution' && $op === 'create' && !empty($params['membership_id']));
+  static $isUpfrontContribution = TRUE;
+  if ($membershipContributionCreation && $isPaymentPlanPayment && $isUpfrontContribution) {
```
MiyaNoctem: I don't think I understand what the purpose of $isUpFrontContribution is. We set it right before we evaluate it, and if the hook is run for a membership creation for a payment plan, then it is then set to false. But it is never checked again. Am I missing something?

omarabuhussein: Sorry, I used the wrong variable names here by mistake, this should be $isFirstPaymentPlanContribution  instead of $isUpFrontContribution (I will fix it in my next commit).

But to explain it, as you know civicrm will be creates the first contribtuion by default but the amount of the contribution and line items will not be correct for it when we pay using payment plan option.
When this hook get called for the first time and the payment is done using payment plan option, then we know for sure that the its get called before creating the first payment plan contribtuion, so we create the recurring contribution and update the first contribution amounts using  :

```php
$paymentPlanProcessor = new CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor($params);
 $paymentPlanProcessor->createPaymentPlan();
```

then I set $isFirstPaymentPlanContribution to false after doing that, because when the rest of the payment plan contributions get created, we don't want to alter their amounts since we create and calculate their amounts correctly in our extension before creating them so altering their amounts is not needed. Hence that this variable is static, so each time the hook is called the previous value will be preserved resulting on only calling this line : 

```php
$paymentPlanProcessor = new CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor($params);
 $paymentPlanProcessor->createPaymentPlan();
```

once and only for the first payment plan contribution.

PR #16, opened 2018-04-29
BASW-104: Adding drupal module to handle the creation of upfront contributions for payment plan submission through webform

omarabuhussein:
## Problem

When submitting a payment plan using the webform using more than one installment, only the recurring contribution will be created as well as the first contribution. But we need all the renaming contributions to be also created upfront.

## Solution

The generating of the renaming upfront contributions should happen after the creation of the first contribution and the recurring contribution, which means after the webform submission, but since it is hard to know that inside the civicrm extension, I had to create a new drupal module called **membershipextras_paymentplan**, this module implements **hook_webform_submission_insert** hook and use reflection to read **wf_crm_webform_postprocess** class private properties and fetch the first contribution ID. If the first contribution ID has a recurring contribution linked to it then we use consider it a payment plan contribution and we call : 

```php
$installmentsHandler = new CRM_MembershipExtras_Service_MembershipInstallmentsHandler($contributionRecurId);
$installmentsHandler->createRemainingInstalmentContributionsUpfront();
```

to generate the remaining contributions upfront.

### DIFF
--- a/CRM/MembershipExtras/Service/MembershipInstallmentsHandler.php
+++ b/CRM/MembershipExtras/Service/MembershipInstallmentsHandler.php
@@ -212,13 +212,7 @@ class CRM_MembershipExtras_Service_MembershipInstallmentsHandler {
       if ($lineItem['entity_table'] === 'civicrm_contribution') {
         $entityID = $contribution->id;
       }
-
-
-      $taxAmount = 0;
-      if (!empty($lineItem['tax_amount'])) {
-        $taxAmount = $lineItem['tax_amount'];
-      }
-
+      
       $lineItemParms = [
         'entity_table' => $lineItem['entity_table'],
         'entity_id' => $entityID,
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Service/MembershipInstallmentsHandler.php
```diff
@@ -213,19 +213,25 @@ private function createLineItems(CRM_Contribute_BAO_Contribution $contribution)
         $entityID = $contribution->id;
       }
 
+
+      $taxAmount = 0;
+      if (!empty($lineItem['tax_amount'])) {
+        $taxAmount = $lineItem['tax_amount'];
```
MiyaNoctem: It seems `$taxAmount` is never used. Can we remove this whole if statement?

CRM/MembershipExtras/Service/MembershipInstallmentsHandler.php
```diff
@@ -213,19 +213,25 @@ private function createLineItems(CRM_Contribute_BAO_Contribution $contribution)
...
         'label' => $lineItem['label'],
         'qty' => $lineItem['qty'],
         'unit_price' => $lineItem['unit_price'],
         'line_total' => $lineItem['line_total'],
-        'price_field_value_id' => $lineItem['price_field_value_id'],
+        'price_field_value_id' => CRM_Utils_Array::value('price_field_value_id', $lineItem),
         'financial_type_id' => $lineItem['financial_type_id'],
         'non_deductible_amount' => $lineItem['non_deductible_amount'],
-        'tax_amount' => $lineItem['tax_amount'],
+        'tax_amount' => CRM_Utils_Array::value('tax_amount', $lineItem),
```
MiyaNoctem: Is there a reason why only some of `$lineItem`'s values are accessed using `CRM_Utils_Array::value()`?

omarabuhussein: yes, if you are paying for a membership along with extra contribution, the extra contribution line ittem will not have all the data available in the membership line item such as the price_set_id since they will be  NULL for the extra contribution line item.

PR #17, opened 2018-04-30
DD-26: Create External ID for Direct Debit Related Entities

zelatyna:
This implements both parts of the ticket:
 1) creating custom fields to support External ID for all 4 entities. Custom fields definitions for Contribution, Recurring Contribution, Memberships are added to customField xml file ,  but for Line Item is done via API
 2) Additional requirements - adding unique constraint for  external ids  on install() and removing fields when extensions is uninstalled 
---
some notes:
- External ID custom fields were supposed to be inactive so both custom group and custom field have property is_active=0. Please advise if it is correct
- Line Item is added as option value to cg_extends_object, but not removed when extension is uninstalled.

### DIFF
--- a/CRM/MembershipExtras/Upgrader.php
+++ b/CRM/MembershipExtras/Upgrader.php
@@ -13,6 +13,7 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
   {
     $this->createOfflineAutoRenewalScheduledJob();
     $this->createLineItemExternalIDCustomField();
+    $this->executeSqlFile('sql/set_unique_external_ids.sql');
   }
 
   /**
@@ -92,7 +93,7 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
       ]);
     }
   }
-
+  
     public function enable() {
     $paymentProcessor = new OfflineRecurringPaymentProcessor();
     $paymentProcessor->toggle(TRUE);
--- a/CRM/MembershipExtras/Upgrader.php
+++ b/CRM/MembershipExtras/Upgrader.php
@@ -9,8 +9,10 @@ use CRM_MembershipExtras_PaymentProcessor_OfflineRecurringContribution as Offlin
  */
 class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
 
-  public function Install() {
+  public function Install()
+  {
     $this->createOfflineAutoRenewalScheduledJob();
+    $this->createLineItemExternalIDCustomField();
   }
 
   /**
@@ -26,6 +28,7 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
       'api_action' => 'run',
     ]);
 
+
     $paymentProcessorType = new ManualRecurringPaymentProcessorType();
     $paymentProcessorType->create();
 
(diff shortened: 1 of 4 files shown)


guanhuan:
@omarabuhussein @MiyaNoctem could one of you review this please? thanks!

CRM/MembershipExtras/Upgrader.php
```diff
@@ -9,8 +9,11 @@
  */
 class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
 
-  public function Install() {
+  public function Install()
+  {
```
MiyaNoctem: Please open the brackets on the same line as the function's name definition.

eg.
```php
public function Install() {
  ...
}
```

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,14 +29,72 @@ private function createOfflineAutoRenewalScheduledJob() {
...
 
     $paymentProcessor = new OfflineRecurringPaymentProcessor();
     $paymentProcessor->create();
   }
 
-  public function enable() {
+    /**
+     * Creates 'External ID' Custom Field for LineItem
+     */
+  private function createLineItemExternalIDCustomField()
```
MiyaNoctem: Please open the brackets on the same line as the function's name definition.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,14 +29,72 @@ private function createOfflineAutoRenewalScheduledJob() {
...
+    if (!$customGroups['count']) {
+      $customGroups = civicrm_api3('CustomGroup', 'create', [
+        'extends' => 'LineItem',
+        'name' => 'line_item_external_id',
+        'title' => E::ts('Line Item External ID'),
+        'table_name' => 'civicrm_value_line_item_ext_id',
+        'is_active' => 0,
+        'style' => 'Inline',
+        'is_multiple' => 0,
+
```
MiyaNoctem: Please remove this blank line.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,14 +29,72 @@ private function createOfflineAutoRenewalScheduledJob() {
...
+        'name' => 'external_id',
+        'label' => E::ts('External ID'),
+        'data_type' => 'String',
+        'html_type' => 'Text',
+        'required' => 0,
+        'is_active' => 0,
+        'is_searchable' => 1,
+        'column_name' => 'external_id',
+        'is_view' => 1
+
```
MiyaNoctem: Please remove this blank line.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,14 +29,72 @@ private function createOfflineAutoRenewalScheduledJob() {
...
+        'is_active' => 0,
+        'style' => 'Inline',
+        'is_multiple' => 0,
+
+      ]);
+    }
+
+    $customFields = civicrm_api3('CustomField', 'get', [
+      'custom_group_id' => $customGroups['id'],
+    ]);
```
MiyaNoctem: You should probably also check for the specific field, not just if the custom group has fields. Just in case.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,14 +29,72 @@ private function createOfflineAutoRenewalScheduledJob() {
...
+        'is_active' => 0,
+        'is_searchable' => 1,
+        'column_name' => 'external_id',
+        'is_view' => 1
+
+      ]);
+    }
+  }
+  
+    public function enable() {
```
MiyaNoctem: There seems to be something wrong with this line's indentation.

MiyaNoctem: There seems to be something wrong with this line's indentation.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -74,7 +135,9 @@ public function uninstall() {
     $paymentProcessorType->remove();
 
     $this->removeOfflineAutoRenewalScheduledJob();
-  }
+    $this->removeCustomExternalIDs();
+
+    }
```
MiyaNoctem: Indentation here also seems inadequate. Should be only 2 spaces right?

CRM/MembershipExtras/Upgrader.php
```diff
@@ -9,8 +9,10 @@
  */
 class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
 
-  public function Install() {
+  public function Install(){
```
MiyaNoctem: Please leave a space between the closing parenthesis and the opening bracket.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,13 +28,70 @@ private function createOfflineAutoRenewalScheduledJob() {
       'api_action' => 'run',
     ]);
 
+
```
MiyaNoctem: This blank line isn't needed.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,13 +28,70 @@ private function createOfflineAutoRenewalScheduledJob() {
...
 
     $paymentProcessor = new OfflineRecurringPaymentProcessor();
     $paymentProcessor->create();
   }
 
+  /**
+  * Creates 'External ID' Custom Field for LineItem
+  */
+  private function createLineItemExternalIDCustomField()
+  {
```
MiyaNoctem: Move this opening bracket to the line where the method is declared.

CRM/MembershipExtras/Upgrader.php
MiyaNoctem: I think this is still missing... it should be:
```php
public function Install() {
  ...
}
```

CRM/MembershipExtras/Upgrader.php
MiyaNoctem: This  bank line is still there and it is not needed, so please remove it.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -26,13 +28,69 @@ private function createOfflineAutoRenewalScheduledJob() {
...
     $paymentProcessorType->create();
 
     $paymentProcessor = new OfflineRecurringPaymentProcessor();
     $paymentProcessor->create();
   }
 
+  /**
+   * Creates 'External ID' Custom Field for LineItem
+   */
+  private function createLineItemExternalIDCustomField(){
```
MiyaNoctem: There should be a space between ')' and '{', ie. `private function functionname() {`.

PR #18, opened 2018-05-03
BASW-106: Add membership offline auto-renew support

omarabuhussein:
## Before

There is no way to create a membership that auto renew automatically from CiviCRM admin panel for offline paid memberships.

## After

A new checkbox added to membership create/renew forms called  "Auto-renew offline?" that allow users to set a membership to be auto-renewed. If this option is already checked, then you cannot check it again when you renew the membership later (so it can only be used once for the membership).

When you check "Auto-renew offline?" option, and If the membership payment selected is "contribution" 
then a new recurring contribution with no installments will be created, if "payment plan" option is selected, then the payment plan recurring contribution will be used for the auto-renewal.


CRM/MembershipExtras/Hook/PostProcess/MembershipOfflineAutoRenewProcessor.php
```diff
@@ -0,0 +1,247 @@
+<?php
+
+class CRM_MembershipExtras_Hook_PostProcess_MembershipOfflineAutoRenewProcessor{
+
+  /**
+   * @var \CRM_Member_Form_Membership
```
MiyaNoctem: Here we also need to update the type.

CRM/MembershipExtras/Hook/PostProcess/MembershipOfflineAutoRenewProcessor.php
```diff
@@ -0,0 +1,247 @@
...
+
+  /**
+   * Creates recurring contribution for the auto-renewed
+   * membership.
+   * (For contribution payment type)
+   *
+   * @return int
+   */
+  private function createAutoRenewRecurContribution() {
+    $PaymentInstrument = 'null';
```
MiyaNoctem: Variable names are supposed to start with lowercase letters by coding standards.

CRM/MembershipExtras/Hook/PostProcess/MembershipOfflineAutoRenewProcessor.php
```diff
@@ -0,0 +1,247 @@
...
+    $formReflection = new ReflectionObject($this->form);
+    $propertyReflection = $formReflection->getProperty($propertyName);
+    $propertyReflection->setAccessible(true);
+    return $propertyReflection->getValue($this->form);
+  }
+
+  /**
+   * Sets the recurring contribution auto_renew
+   * field to true.
+   *
```
MiyaNoctem: Please remove extra blank line.

PR #20, opened 2018-05-09
BASW-91: Add Setting to Choose Default Payment Processor

MiyaNoctem:
## Overview
We need an option on *Administer > Payment Plan Settings*, to select the payment processor that should be used when creating payment plans to pay for memberships. This new setting should:

1. Be called "Offline payment processor for back office" with helptext "Select the payment processor that should be used when creating payment plans via CiviCRM admin interface such as new/ renew membership.".
2. Show a list containing all payment processors that use a payment processor type of "Payment_Manual" payment class.
3. Be required and should default to the offline payment processor this extension creates.
4. When creating a payment plan using new/renew membership admin interface, the id of the payment processor selected in the setting should be used as the recurring contribution payment processor id.

## Before
Payment plans were created with the hard-coded payment processor this extension creates.

## After
Added a new setting to choose payment processor from all processors of Manual Payment type. This processor is then used when creating and renewing memberships with a payment plan.


CRM/MembershipExtras/Form/PaymentPlanSettings.php
```diff
@@ -24,12 +26,28 @@ public function buildQuickForm() {
 
     $settingFields  = $this->getSettingFields();
     $settingFieldNames = [];
+
     foreach ($settingFields  as $name => $field) {
+      $attributes = '';
+
+      switch (true) {
```
omarabuhussein: I always see that such kind of switch ins unhealthy to have in code and does not make any sense, it is better to be replace by if statements in this case, or probably it is better now to remove the whole foreach loop and adding fields one by one since with this new field it does not make much sense to still using the loop.

CRM/MembershipExtras/Form/PaymentPlanSettings.php
```diff
@@ -52,6 +70,15 @@ public function buildQuickForm() {
     $this->assign('settingFields', $settingFieldNames);
   }
 
+  /**
+   * Builds an array mapping manual payment processor id's to processor name.
+   *
+   * @return array
+   */
+  private function getManualPaymentProcessors() {
+    return CRM_MembershipExtras_Service_ManualPaymentProcessors::getProcessorsIDNameMap();
```
omarabuhussein: why to wrap this into a new method where u can just call the method directly above, since you are not doing any thing else other than calling getProcessorsIDNameMap() method.

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -192,14 +192,13 @@ private function getOfflineAutoRenewalRecurContributions() {
...
     $query = 'SELECT ccr.id as contribution_recur_id, ccr.installments  
-              FROM civicrm_contribution_recur ccr 
-              LEFT JOIN  civicrm_membership cm 
-                ON ccr.id = cm.contribution_recur_id 
-              WHERE ccr.auto_renew = 1 
-                AND (ccr.payment_processor_id IS NULL OR ccr.payment_processor_id IN (' . $manualPaymentProcessorsIDs . '))
-                AND (ccr.contribution_status_id != ' . $cancelledStatusID . ' OR  ccr.contribution_status_id != ' . $refundedStatusID . ')
-                AND cm.end_date <= CURDATE() 
-              GROUP BY ccr.id';
+                FROM civicrm_contribution_recur ccr
```
omarabuhussein: did u changed anything in this query ? or u just changed the format ?

CRM/MembershipExtras/PaymentProcessor/OfflineRecurringContribution.php
```diff
@@ -41,6 +41,12 @@ public function create() {
     ];
 
     $paymentProcessor = civicrm_api3('PaymentProcessor', 'create', $params);
+    civicrm_api3(
```
omarabuhussein: please add space before this line

CRM/MembershipExtras/Service/ManualPaymentProcessors.php
```diff
@@ -24,4 +23,26 @@ public static function getIDs() {
     return $manualPaymentProcessorsList;
   }
 
+  /**
+   * Builds an array mapping manual payment processor id's to processor name.
+   *
+   * @return array
+   */
+  public static function getProcessorsIDNameMap() {
```
omarabuhussein: **getIDNameMap** would be enough

CRM/MembershipExtras/PaymentProcessor/OfflineRecurringContribution.php
```diff
@@ -63,6 +69,30 @@ public function get() {
     return $processor['values'][0];
   }
 
+  /**
+   * Returns the details of the default payment processor as per payment plan
+   * settings, or NULL if it does not exist.
+   *
+   * @return array
+   */
+  public function getDefaultProcessor() {
```
omarabuhussein: I don't think method belongs to this class, it has nothing to do with. Maybe u can create a new class called CRM_MembershipExtras_Settings and move this method to it as a static method.

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
MiyaNoctem: I had originally, but it was basically the same change you did on BASW-68, so I left your version when I solved the conflict. So, to answer your question, yes, now the change is only in format.

CRM/MembershipExtras/SettingsManager.php
```diff
@@ -0,0 +1,32 @@
...
+ */
+class CRM_MembershipExtras_SettingsManager {
+
+  /**
+   * Returns the details of the default payment processor as per payment plan
+   * settings, or NULL if it does not exist.
+   *
+   * @return array
+   */
+  public static function getDefaultProcessor() {
```
omarabuhussein: I think it is better to rename this to getDefaultProcessorId() and only return the ID, if we needed more info later then we can change it but for now lets just stick to the ID.

PR #21, opened 2018-05-16
BASW-90: Fixing line item amounts and membership end date when renewing memberships

omarabuhussein:
## Before
When renewing a membership using "payment plan" option and selecting the payment status as "pending", the line item amounts will not be correct and the membership will not get extended.

The membership extending part is how civicrm core work, it does not extend the membership unless u complete the related membership payment but still for this extension we should extend it when paying via payment plan. The extending should be similar to how it work in offline renewal scheduled job.

## After

Now the line items amounts are correct and the membership get extended as soon as u submit the renewal form.

sssss

## Technical notes 

- The logic that calculates the end date for extending the membership is moved to a new class called : **CRM_MembershipExtras_Service_MembershipEndDateCalculator** , this class is now used in hooks when renewing the membership and for offline renewal scheduled job.

- The extending at renewal only work for payment plan payments and only when the payment status is "pending"

- The reason for line item amount problem, is that the financial item creation order is different from membership creation. For example at membership creation, the MembershipPayment record is created before LineItem record, but for renewal the MembershipPayment is created after the LineItem, So the logic implemented in pre hook is not valid anymore since it was depend on the membership creation order and I had to change it so it does not care about MembershipPayment  record anymore.

### DIFF
--- a/CRM/MembershipExtras/Hook/PreEdit/Membership.php
+++ b/CRM/MembershipExtras/Hook/PreEdit/Membership.php
@@ -1,5 +1,7 @@
 <?php
 
+use CRM_MembershipExtras_Service_MembershipEndDateCalculator as MembershipEndDateCalculator;
+
 /**
  * Called by membershipextras_civicrm_pre hook
  * before editing/renewing a membership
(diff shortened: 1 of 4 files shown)


omarabuhussein:
since I don't have much time and more work is depending on this and no one can review currently I will have to merge this without code review.

PR #23, opened 2018-05-16
BASW-113: Cancel Existing Payment Plan

MiyaNoctem:
## Overview
When cancelling a recurring contribution, a popup will show up and ask for confirmation. A new dialog box containing two checkbox options should be used to replace the popup.

The first checkbox is "Do you wish to cancel any pending instalment contribution?" which if checked and confirmed, as a result, all "Pending" contributions that are linked to the recurring contribution will also be cancelled.

The second checkbox is "Do you wish to cancel any linked membership?" which if checked and confirmed, as a result, all memberships that are linked to the recurring contribution will also be cancelled.

## Before
Enable/Disable CiviCRM Page was being used to handle recurring contribution cancellation.

image

## After
Altered the cancellation link on recurring contributions list so it uses a new form, with the required checkboxes. If they are clicked, pending contributions and/or related memberships are cancelled too.

image

### DIFF
--- a/CRM/MembershipExtras/Form/CancelRecurringContribution.php
+++ b/CRM/MembershipExtras/Form/CancelRecurringContribution.php
@@ -108,7 +108,9 @@ class CRM_MembershipExtras_Form_CancelRecurringContribution extends CRM_Core_For
       'options' => ['limit' => 0],
       'api.Contribution.create' => array(
         'id' => '$value.id',
-        'contribution_status_id' => 'Cancelled'
+        'contribution_status_id' => 'Cancelled',
+        'cancel_date' => date('Y-m-d H:i:s'),
+        'cancel_reason' => 'Cancelled because related recurring contribution was cancelled.',
       ),
     ]);
   }
--- a/CRM/MembershipExtras/Form/CancelRecurringContribution.php
+++ b/CRM/MembershipExtras/Form/CancelRecurringContribution.php
@@ -1,0 +1,125 @@
+<?php
+
+/**
+ * QuickForm used to cancel memberships.
+ */
+class CRM_MembershipExtras_Form_CancelRecurringContribution extends CRM_Core_Form {
+
+  /**
+   * ID of recurring contribution to be cancelled.
+   *
+   * @var int
+   */
+  private $id;
+
+  /**
+   * ID of contact from where contributions is being cancellsd.
+   *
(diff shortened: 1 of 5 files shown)


CRM/MembershipExtras/Form/CancelRecurringContribution.php
```diff
@@ -0,0 +1,127 @@
...
+        'status_id' => 'Cancelled',
+      ],
+    ]);
+  }
+
+  /**
+   * Cancels pending contributions associated to current recurring contribution.
+   */
+  private function cancelPendingInstallments() {
+    civicrm_api3('Contribution', 'get', [
```
omarabuhussein: I am wondering, is canceling a contribution only change the contribution status (alongside updating the cancle_date and cancle_reason) or is there maybe some other stuff  that civi do in the background ? If you did not do that already, I think it is better to check the queries happen when canceling a membership and ensure that it is similar to what you did below (you can use the tip I shared in email last week to check all queries happen during a certain operation in case u don't know it already).

CRM/MembershipExtras/Hook/Links/RecurringContribution.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+class CRM_MembershipExtras_Hook_Links_RecurringContribution {
+
+  /**
+   * Alters 'Cancel' action link to use a custom form, instead of CiviCRM's
+   * default enable/disable weird control.
+   *
+   * @param array $links
+   */
+  public function alterLinks(&$links) {
```
omarabuhussein: maybe it is better to pass the links to the constructor instead.

CRM/MembershipExtras/Form/CancelRecurringContribution.php
omarabuhussein: I see, and since you are using cancel API action on ContributionRecur endpoint it will take care of the stuff above. but there is something that we might might need to ask @guanhuan about which is : 

so If I cancel a payment plan (recurring contribution) that has 12 installments attached to it, should the records above get generated only for the recurring contribution or should they get generated also for all the payment plan contributions ?

MiyaNoctem: The records above are for payment plan installments exclusively.

PR #27, opened 2018-05-21
BASW-90: Fix Membership Creation and Renewal Forms Validation

MiyaNoctem:
## Overview
By default, payment plan fields are set as Number of Instalments = 12, Interval=1 , Unit = month now. But if someone removes the Number of Instalments and Interval fields and save there is no validation. Also, there should be a red asterisk next to both fields, so it is clear both are required.

## Before
Validation was not being executed for for empty installments and frequencies. There was no visual marker that obviously stated the payment plan fields were required.

## After
Validation was only run if number of installments was > 1. Renamed method that checked if a payment plan is being used so it is clear it also validates if more than one contribution is being used to pay. Added validation of number of installments. Added red *'s to show fields 'installments', 'installments_frequency', and 'installments_frequency_unit' are required.

### DIFF
--- a/templates/CRM/Member/Form/PaymentPlanToggler.tpl
+++ b/templates/CRM/Member/Form/PaymentPlanToggler.tpl
@@ -104,12 +104,14 @@
     </td>
   </tr>
   <tr id="installments_row">
-    <td nowrap>{$form.installments.label}</td>
+    <td nowrap>
+      {$form.installments.label} <span class="marker">*</span>
+    </td>
     <td nowrap>
       {$form.installments.html}
-      {$form.installments_frequency.label}
+      {$form.installments_frequency.label} <span class="marker">*</span>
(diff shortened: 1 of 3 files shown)


templates/CRM/Member/Form/PaymentPlanToggler.tpl
```diff
@@ -104,12 +104,14 @@
...
   <tr id="installments_row">
-    <td nowrap>{$form.installments.label}</td>
+    <td nowrap>
+      {$form.installments.label} <span class="marker">*</span>
+    </td>
     <td nowrap>
       {$form.installments.html}
-      {$form.installments_frequency.label}
+      {$form.installments_frequency.label} <span class="marker">*</span>
       {$form.installments_frequency.html}
```
omarabuhussein: installments_frequency is also a required field for payment plan.

PR #28, opened 2018-05-28
BASW-118: Update Installments on Recurring Contribution Update

MiyaNoctem:
## Overview
We need to alter Update Recurring contribution form to expose some new fields and allow users to update related pending installments if they choose to do so. For this, we need to:

- Make "Number of instalment" field readonly if the recurring contribution is using a payment processor with "Payment_Manual" class.
- Expose the "Auto Renew?" field on "Update Recurring Contribution" form.
- Expose "Payment Method" field, "Cycle Day" field on "Update Recurring Contribution" form.
- When admin click on save on "Update Recurring Contribution" form, if "Payment Mehod" or "Cycle Day" is changed, a popup should appear asking "Do you want to update any outstanding instalment contribution with the new Payment Method/ Cycle Day?".
- If admin chose "No" in the previous popup. No existing contribution will get affected.
- If admin chose "Yes" in the previous popup. Any existing Pending contribution under the recurring contribution will reflect the new "Payment Method" and "Cycle Day".

## Before
image

## After
Added cycle_day and payment_instrument_id fields to Update Recurring Contribution Form, and added a confirmation dialog so that if user chooses to, pending installments' receive date and payment instrument are also updated.

basw-118-after

### DIFF
--- a/CRM/MembershipExtras/Hook/BuildForm/UpdateSubscription.php
+++ b/CRM/MembershipExtras/Hook/BuildForm/UpdateSubscription.php
@@ -1,0 +1,84 @@
+<?php
+
+/**
+ * Alters UpdateSubscription form.
+ */
+class CRM_MembershipExtras_Hook_BuildForm_UpdateSubscription {
+
+  private $form;
+  private $templatePath;
+  private $recurringContribution;
+
+  public function __construct(CRM_Contribute_Form_UpdateSubscription $form) {
+    $this->form = $form;
+    $this->templatePath = CRM_MembershipExtras_ExtensionUtil::path() . '/templates';
+    $this->loadRecurringContributionData();
+  }
+
+  /**
+   * Loads data for the current recurring contribution.
+   */
+  private function loadRecurringContributionData() {
+    $recurringContributionID = CRM_Utils_Request::retrieve('crid', 'Integer', $this->form, FALSE);
+
+    $this->recurringContribution = civicrm_api3('ContributionRecur', 'get', [
+      'sequential' => 1,
+      'id' => $recurringContributionID,
+    ])['values'][0];
+  }
+
+  /**
+   * Implements modifications to UpdateSubscription form.
+   */
+  public function buildForm() {
+    if ($this->isManualPaymentPlan()) {
+      $amount = $this->form->getElement('amount');
+      $amount->setAttribute('readonly', true);
+
+      $installments = $this->form->getElement('installments');
+      $installments->setAttribute('readonly', true);
+
+      $this->form->add('checkbox', 'auto_renew', ts('Auto-renew?'));
+      $this->form->setDefaults(['auto_renew' => $this->recurringContribution['auto_renew']]);
+
+      $this->form->add('select', 'payment_instrument_id',
+        ts('Payment Method'),
+        ['' => ts('- select -')] + CRM_Contribute_PseudoConstant::paymentInstrument(),
+        TRUE
+      );
+      $this->form->setDefaults(['payment_instrument_id' => $this->recurringContribution['payment_instrument_id']]);
+      $this->form->assign('isBackOffice', 1);
+
+      $this->form->add('text', 'cycle_day', ts('Cycle Day'), TRUE);
+      $this->form->setDefaults(['cycle_day' => $this->recurringContribution['cycle_day']]);
+
+      CRM_Core_Region::instance('page-body')->add([
+        'template' => "{$this->templatePath}/CRM/Member/Form/UpdateSubscriptionModifications.tpl"
+      ]);
+
+      $this->form->addButtons([
+        [
+          'type' => 'upload',
+          'name' => ts('Confirm'),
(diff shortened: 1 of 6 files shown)


CRM/MembershipExtras/Hook/BuildForm/UpdateSubscription.php
```diff
@@ -0,0 +1,85 @@
+<?php
+
+/**
+ * Alters UpdateSubscription form.
+ */
+class CRM_MembershipExtras_Hook_BuildForm_UpdateSubscription {
+
+  private $form;
```
omarabuhussein: please add docblocks to these properties.

CRM/MembershipExtras/Hook/BuildForm/UpdateSubscription.php
```diff
@@ -0,0 +1,85 @@
...
+
+  private $form;
+  private $templatePath;
+  private $recurringContribution;
+
+  public function __construct(CRM_Contribute_Form_UpdateSubscription $form) {
+    $this->form = $form;
+    $this->templatePath = CRM_MembershipExtras_ExtensionUtil::path() . '/templates';
+
+    $recurringContributionID = CRM_Utils_Request::retrieve('crid', 'Integer', $this->form, FALSE);
```
omarabuhussein: you can move the following two lines to a new method called : 

$this->setRecurringContribution()

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -0,0 +1,92 @@
...
+  /**
+   * Form to be processed.
+   *
+   * @var \CRM_Contribute_Form_UpdateSubscription
+   */
+  private $form;
+
+  public function __construct(CRM_Contribute_Form_UpdateSubscription $form) {
+    $this->form = $form;
+    $this->recurringContributionID = CRM_Utils_Request::retrieve('crid', 'Integer', $this->form, TRUE);
```
omarabuhussein: same here, the following two lines can be moved to new method called $this->setRecurringContribution().

Also no need to store both recurringContributionID  and recurringContribution properties, the latter is enough and you can access the ID using $this->recurringContribution['id']

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  /**
+   * Postprocesses CRM_Contribute_Form_UpdateSubscription to deal with added
+   * fields.
+   */
+  public function postProcess() {
+    $updateInstallments = CRM_Utils_Request::retrieve('update_installments', 'Integer', $this->form, FALSE);
+
+    if ($updateInstallments == '1') {
```
omarabuhussein: I think using guard statement is better here : 

if (!$updateInstallments) {
return;
}

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -0,0 +1,92 @@
...
+      $params = $this->form->exportValues();
```
omarabuhussein: I think it better to compare the cycle date in $params with the cycle date in $this->recuringContribution if it changed or not, if not then there is no need to recalculate the receive date.

CRM/MembershipExtras/Hook/BuildForm/UpdateSubscription.php
```diff
@@ -5,15 +5,39 @@
...
+   * Array with the data of the recurring contribution that is being updated.
+   *
+   * @var array
+   */
   private $recurringContribution;
 
   public function __construct(CRM_Contribute_Form_UpdateSubscription $form) {
     $this->form = $form;
     $this->templatePath = CRM_MembershipExtras_ExtensionUtil::path() . '/templates';
+    $this->loadRecurringContribution();
```
omarabuhussein: I think setRecurringContribution is a better name

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -14,54 +14,94 @@ class CRM_MembershipExtras_Hook_PostProcess_UpdateSubscription {
...
+  private $receiveDateCalculator;
+
   public function __construct(CRM_Contribute_Form_UpdateSubscription $form) {
     $this->form = $form;
-    $this->recurringContributionID = CRM_Utils_Request::retrieve('crid', 'Integer', $this->form, TRUE);
-    $this->recurringContribution = civicrm_api3('ContributionRecur', 'getsingle', [
-      'id' => $this->recurringContributionID
-    ]);
-    $this->receiveDateCalculator = new InstallmentReceiveDateCalculator($this->recurringContribution);
+    $this->loadRecurringContribution();
```
omarabuhussein: same here, setRecurringContribution

PR #29, opened 2018-05-30
BASW-122: Change payment plan recurring contribution status when paying installments

omarabuhussein:
## Before

When paying the first payment plan installment (completely or partially) or when the all the payment plan installments are completed, the recurring contribution status will stay on "pending" status.

ddsadas

## After

- When paying the first payment plan installment (completely or partially) , the recurring contribution status will change to "in progress".

- When paying completing all the payment plan installments, the recurring contribution will change to "Completed".

before

## Other Notes

There was a problem where the membership get extended when you finish the last payment, I fixed it in this commit : 

https://github.com/compucorp/uk.co.compucorp.membershipextras/commit/3fff6ec4d9d2441d2c15ae9acf9e8617a87ba254

## Limitations

Currently there is a problem with handling "partial payments" that emerge from CiviCRM side and hard to be solved currently (without probably some fixes to CiviCRM core) , the problem consist of two parts : 

1- If you paid the **last** installment partially then you completed it, then the recurring contribution status will stay "in progress" instead of "completed".

2- If for partially paid any installment (other than the first installment) then completed it, then the membership will get extended when you complete the partially paid installment.

Here is a gif that demonstrate the two issues : 

aaaa

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/Contribution.php
+++ b/CRM/MembershipExtras/Hook/Pre/Contribution.php
@@ -1,0 +1,108 @@
+<?php
+
+class CRM_MembershipExtras_Hook_Pre_Contribution {
+
+
+  private $id;
+
+  /**
+   * The contribution to-be-created parameters passed from the hook.
+   *
+   * @var array
+   */
+  private $params;
+
+  /**
+   * The payment plan recurring contribution data if exist
+   *
+   * @var array|null
+   */
+  private $recurContribution = NULL;
+
+  /**
+   * Is this an offline payment plan payment ?
+   *
+   * @var bool
+   */
+  private $isOfflinePaymentPlanPayment = FALSE;
+
+  public function __construct($id, &$params) {
+    $this->id = $id;
+    $this->params = $params;
+    $this->setRecurContribution();
+    $this->setIsOfflinePaymentPlanPayment();
+  }
+
+  /**
+   * Sets $recurContribution
+   */
+  private function setRecurContribution() {
+    if (empty($this->params['prevContribution']->contribution_recur_id)) {
+      return;
+    }
+
+    $this->recurContribution = civicrm_api3('ContributionRecur', 'get', [
+      'sequential' => 1,
+      'id' => $this->params['prevContribution']->contribution_recur_id,
+    ])['values'][0];
+  }
+
+  /**
+   * Sets $isOfflinePaymentPlanPayment
+   */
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/Pre/Contribution.php
```diff
@@ -0,0 +1,108 @@
...
+    $partiallyPaidInstallmentsCount = civicrm_api3('Contribution', 'getcount', [
+      'contribution_recur_id' => $this->recurContribution['id'],
+      'contribution_status_id' => 'Partially paid',
+    ]);
+
+    $newStatus = NULL;
+    if (($partiallyPaidInstallmentsCount == 1 && $paidInstallmentsCount == 0) ||
+      ($partiallyPaidInstallmentsCount == 0 && $paidInstallmentsCount == 1) ||
+      ($partiallyPaidInstallmentsCount == 0 && $paidInstallmentsCount == 0)
+    ) {
```
MiyaNoctem: Shouldn't the recurring contribution be 'In Progress' if either of those values is more than zero and paid installments is less than the number of installments? Although, since 'Completed' status is taken care of on the next if statement, it might be enough just to check if either of those values is more than zero. Also, I don't really understand why if both are zero, the recurring contribution would be In progress... shouldn't it stay 'Pending' if no payments have started?

Eg.
```php
if ($partiallyPaidInstallmentsCount > 0 || $paidInstallmentsCount > 0) {
  $newStatus = 'In Progress';
}
```

membershipextras.php
```diff
@@ -157,13 +157,19 @@ function membershipextras_civicrm_pre($op, $objectName, $id, &$params) {
    * options for now.
    */
   static $contributionID = NULL;
+  static $preEditContributionHook = NULL;
   if ($op === 'edit' && $objectName === 'Contribution') {
     $contributionID = $id;
+
+    $preEditContributionHook = new CRM_MembershipExtras_Hook_Pre_Contribution($id, $params);
+    $preEditContributionHook->updatePaymentPlanStatus();
```
MiyaNoctem: Why did  you choose to update the recurring contribution's status before the contribution is saved? Wouldn't it make more sense to do it after, as in the contribution's post hook?

omarabuhussein:
@MiyaNoctem 

Hey man, thanks for your comments, they made me to rethink my approach again and I opened a new PR that implement this differently, the new PR also fix the first  mentioned limitation in this PR.

https://github.com/compucorp/uk.co.compucorp.membershipextras/

PR #33, opened 2018-06-04
BASW-123: Prevent Membership cancellation on payment plan cancellation

omarabuhussein:
This PR is based on the new hook added in this Core PR : https://github.com/civicrm/civicrm-core/pull/12253 which is not yet reviewed/merged.


omarabuhussein:
Update: Since we are using a forked CiviCRM version currently which have the core PR above merged to it, I will merged this PR.

PR #38, opened 2018-06-15
BASW-91: Show if Payment Processors are Live or Not on Settings

MiyaNoctem:
## Overview
Manually adding a new payment processor creates two different items, with the same name, one with is_test = 0, the other one with is_test = 1. This made it impossible to know which was which on the extension's settings form, where we were just showing the processor's name.

## Before
Settings page for the extension only showed processor's name on the selection bx used to define default payment processor.

## After
Added 'Live - ' and 'Test - ' prefixes correspondingly to each processor, so users can know which is which.

### DIFF
--- a/CRM/MembershipExtras/Form/PaymentPlanSettings.php
+++ b/CRM/MembershipExtras/Form/PaymentPlanSettings.php
@@ -57,7 +57,7 @@ class CRM_MembershipExtras_Form_PaymentPlanSettings extends CRM_Core_Form {
    * @param array $defaultProcessorField
    */
   private function addDefaultProcessorField($defaultProcessorField) {
-    $processorOptions = array('' => ts('- select -')) + ManualPaymentProcessors::getIDNameTestMap();
+    $processorOptions = array('' => ts('- select -')) + ManualPaymentProcessors::getIDNameMap();
 
     $this->add(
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Form/PaymentPlanSettings.php
```diff
@@ -57,7 +57,7 @@ public function buildQuickForm() {
    * @param array $defaultProcessorField
    */
   private function addDefaultProcessorField($defaultProcessorField) {
-    $processorOptions = array('' => ts('- select -')) + ManualPaymentProcessors::getIDNameMap();
+    $processorOptions = array('' => ts('- select -')) + ManualPaymentProcessors::getIDNameTestMap();
```
omarabuhussein: it is more clear to keep the method name to what it was before.

CRM/MembershipExtras/Service/ManualPaymentProcessors.php
```diff
@@ -24,11 +24,12 @@ public static function getIDs() {
   }
 
   /**
-   * Builds an array mapping manual payment processor id's to processor name.
+   * Builds an array mapping manual payment processor id's to processor name and
```
omarabuhussein: the previous description is better.

PR #39, opened 2018-06-25
BASW-175: Unlock the Amount Field for Payment Plan when no Priceset is Used

MiyaNoctem:
## Overview
We want to allow the staff to edit the amount for a payment plan when the following criteria are met:

1. the staff is creating new membership using payment plan or renewing a membership using payment plan
2. priceset option is not used

The final value in the amount field will be the amount for the payment plan. It will be divided by the number of instalments and then used as the instalment contribution amount and the line amount.

## Before
Total amount field was being made read-only if payment plan was used to paid for a membership.

## After
Total amount field is set to readonly if priceset is used to pay for the membership exclusively.

Also fixed a problem with updating amount of first invoice when using price set.

### DIFF
--- a/membershipextras.php
+++ b/membershipextras.php
@@ -300,3 +300,16 @@ function membershipextras_civicrm_links($op, $objectName, $objectId, &$links, &$
     $recurContribuLinksHook->alterLinks();
   }
 }
+
+function membershipextras_civicrm_alterContent(&$content, $context, $tplName, &$object) {
+  $snippet = CRM_Utils_Request::retrieve('snippet', 'Int');
+  $priceSetID = CRM_Utils_Request::retrieve('priceSetId', 'Int');
+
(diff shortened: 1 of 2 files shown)


membershipextras.php
```diff
@@ -300,3 +300,16 @@ function membershipextras_civicrm_links($op, $objectName, $objectId, &$links, &$
     $recurContribuLinksHook->alterLinks();
   }
 }
+
+function membershipextras_civicrm_alterContent(&$content, $context, $tplName, &$object) {
+  $snippet = CRM_Utils_Request::retrieve('snippet', 'Int');
+  $priceSetID = CRM_Utils_Request::retrieve('priceSetId', 'Int');
+
+  if ($tplName == 'CRM/Member/Page/Tab.tpl' && $snippet == 4 && !empty($priceSetID)) {
```
omarabuhussein: replace 4 with a variable where the variable name explain what 4 means here.

PR #40, opened 2018-07-03
BASW-175: Fix Problems Calculating Tax Amounts

MiyaNoctem:
## Overview
Tax amounts used to create contributions and line items when using an edited
total amount were being saved with strange values seemingly not related with the amount given by the user.

## Before
Tax amounts used to create contributions and line items was using the original amount for the membership instead of the edited value, causing problems like negative line item totals and over-priced tax amounts.

## After
Fixed by implementing a Pre hook on Membership entity that recalculates tax amounts using the actual total amount given by the user. Refactored other Membership entity pre hooks into the same class to clean-up the hook's code.


CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor.php
```diff
@@ -114,10 +114,24 @@ private function alterFirstContributionParameters() {
     $this->params['net_amount'] =  $this->recurringContribution['amount'];
 
     if (!empty($this->params['tax_amount'])) {
-      $this->params['tax_amount'] = $this->calculateSingleInstallmentAmount($this->params['tax_amount']);
+      $this->params['tax_amount'] = $this->calculateInstallmentTax($this->params['total_amount']);
```
omarabuhussein: this code works for both price-set and financial type membership payments, doing this will screw priceset payments, so you need to do this only if it this not a price-set payment.

CRM/MembershipExtras/Hook/Pre/Membership.php
```diff
@@ -0,0 +1,130 @@
...
+    $this->id = $id;
+    $this->params = &$params;
+    $this->contributionID = $contributionID;
+  }
+
+  /**
+   * Preprocesses parameters used for Membership operations.
+   */
+  public function preProcess() {
+    switch ($this->op) {
```
omarabuhussein: I don't think combinint the both in one class (CRM_MembershipExtras_Hook_Pre_Membership ) is the best course of action here, maybe just seperate them into : 

CRM_MembershipExtras_Hook_Pre_MembershipEdit 
CRM_MembershipExtras_Hook_Pre_MembershipCreate

if you did this then you can then git rid of the "now" confusing CRM_MembershipExtras_Hook_PreEdit_Membership by combing it into CRM_MembershipExtras_Hook_Pre_MembershipEdit

CRM/MembershipExtras/Hook/Pre/Membership.php
```diff
@@ -0,0 +1,130 @@
...
+        $this->preProcessEdit();
+        break;
+    }
+  }
+
+  /**
+   * Fixes parameters used to create a membership.
+   */
+  private function preProcessCreation() {
+    $this->fixTaxAmount();
```
omarabuhussein: recalculateTaxAmount is probably a more clear name

CRM/MembershipExtras/Hook/Pre/Membership.php
```diff
@@ -0,0 +1,130 @@
...
+    $this->fixLineItems();
```
omarabuhussein: recalculateLineItemsAmounts is probably a more clear name

CRM/MembershipExtras/Hook/Pre/Membership.php
```diff
@@ -0,0 +1,130 @@
...
+      case 'edit':
+        $this->preProcessEdit();
+        break;
+    }
+  }
+
+  /**
+   * Fixes parameters used to create a membership.
+   */
+  private function preProcessCreation() {
```
omarabuhussein: again, these calculations will probably screw the price-set payments, it should only be applied to financial type payments.

CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor.php
MiyaNoctem: Well I tested it with a Priceset and it all seemed ok, but I'll look into it a little deeper just to be safe.

MiyaNoctem: Yeah, it all works correctly. Basically it just recalculates tax amount from the actual value and tax rate, which I think is a good thing, both for the contribution and for each line item in the price set.

image

CRM/MembershipExtras/Hook/Pre/Membership.php
MiyaNoctem: Same, tested and all works ok.

membershipextras.php
```diff
@@ -302,15 +302,8 @@ function membershipextras_civicrm_links($op, $objectName, $objectId, &$links, &$
...
-
-  if ($tplName == 'CRM/Member/Page/Tab.tpl' && $snippet == CRM_Core_Smarty::PRINT_NOFORM && !empty($priceSetID)) {
-    // Throw change event when total_amount is altered
-    $content = preg_replace(
-      '/cj\(\'#total_amount\'\)\.val\((.+)\);/',
-      'cj(\'#total_amount\').val(${1}).change();',
-      $content
-    );
+  if ($tplName == 'CRM/Member/Page/Tab.tpl') {
+    $memberTabPage  = new CRM_MembershipExtras_Hook_Alter_MemberTabPageContent($content);
```
omarabuhussein: the hook name is alterContent , so I think Hook_Alter part in the name of the class should reflect that so  should be renamed to Hook_AlterContent, the whole name will just be : 

CRM_MembershipExtras_Hook_AlterContent_MemberTabPage

CRM/MembershipExtras/Hook/Alter/MemberTabPageContent.php
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Appends a '.change()' to throw the event when the total amount is altered.
+   */
+  private function appendJSCodeToWatchTotalAmountValueChanges() {
+    $snippet = CRM_Utils_Request::retrieve('snippet', 'Int');
+    $priceSetID = CRM_Utils_Request::retrieve('priceSetId', 'Int');
+
+    if ($snippet == CRM_Core_Smarty::PRINT_NOFORM && !empty($priceSetID)) {
+      // Throw change event when total_amount is altered
```
omarabuhussein: no need for this inline comment anymore, the method name already reflect what it does.

CRM/MembershipExtras/Hook/Alter/MemberTabPageContent.php
```diff
@@ -0,0 +1,36 @@
...
+
+  /**
+   * Executes alterations for the content of the page.
+   */
+  public function alterContent() {
+    $this->appendJSCodeToWatchTotalAmountValueChanges();
+  }
+
+  /**
+   * Appends a '.change()' to throw the event when the total amount is altered.
```
omarabuhussein: the docblock contain and implementation detail (Appends a '.change()')  which is not a good idea (because if you changed the method implementation to not use .change() hen you also should remember to change the docblock (and u will probably forget that and the doblock will be outdated) ... I also think in this case that the method name is clear and no need for the docblock at all so u can remove it.

CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor.php
omarabuhussein: it is a problematic  because of this : 

```
    $taxRates = CRM_Core_PseudoConstant::getTaxRates();
    $rate = CRM_Utils_Array::value($this->params['financial_type_id'], $taxRates, 0);
```

you basically use the contribution financial_type_id which is something does not exist when paying using price sets, assume you have 3 line items (price fields) in the price set where each one has a different financial type (and each financial type has a different amount of tax),  but code above will use single tax rate for all the line items (price fields) even if they have different amount of taxes, or at least this is what I can tell from reading the code.

CRM/MembershipExtras/Hook/Pre/MembershipCreate.php
```diff
@@ -0,0 +1,54 @@
...
+
+  /**
+   * Recalculates tax amount tax rate according to selected financial type, as
+   * this can be broken if paying using payment plan or altering total value.
+   */
+  private function recalculateTaxAmount() {
+    $taxRates = CRM_Core_PseudoConstant::getTaxRates();
+    $rate = CRM_Utils_Array::value($this->params['financial_type_id'], $taxRates, 0);
+
+    $this->params['tax_amount'] = ($this->params['total_amount'] * ($rate / 100)) / (1 + ($rate / 100));
```
omarabuhussein: I think you need to use : 

```php
round(AMOUNT, 2)
```
for all amount calculations here so it match how civicrm do it

CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
```diff
@@ -1,149 +1,179 @@
...
+   * Preprocesses parameters used for Membership operations.
+   */
+  public function preProcess() {
+    if ($this->paymentContributionID) {
+      $this->preventExtendingPaymentPlanMembership();
+    }
+
+    $isPaymentPlanPayment = $this->isPaymentPlanWithMoreThanOneInstallment();
+    $isMembershipRenewal = CRM_Utils_Request::retrieve('action', 'String') & CRM_Core_Action::RENEW;
+
```
omarabuhussein: no need for this space

CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor.php
```diff
@@ -113,11 +113,29 @@ private function alterFirstContributionParameters() {
     $this->params['total_amount'] =  $this->recurringContribution['amount'];
     $this->params['net_amount'] =  $this->recurringContribution['amount'];
 
-    if (!empty($this->params['tax_amount'])) {
+    $lineItemCount = $this->countLineITems();
```
omarabuhussein: I don't  think that this is a good solution, because you can have a price set with a single line item, maybe there is some parameters that will be passed (like price set id) that are better suited to detect if this a price set of finical type contribution ?

PR #45, opened 2018-07-24
BASW-211: Copy custom groups during offline auto-renew and add settin to allow for exculding some custom groups

omarabuhussein:
## Before

When the the offline auto-renewal scheduled job auto-renew the offline memberships, the new contributions and the new recurring contribution will not have the custom fields exsisted in the previous contributions and recurring contribution, but we need all of them to be copied, we also want a setting that prevent certain custom groups fields to be copied such as odoo sync information custom group.

2018-07-24 19_18_58-payment plan settings _ dmaster14

## After

The offline auto-renewal scheduled job now copy all the contributions and recurring contribution custom fields from the previous ones. it also has a new multi-select box setting that allow users to choose which custom groups to exclude during the copy.

aaaaaa

## Other notes

I also changed the signature of : 

membershipextras_postOfflineAutoRenewal

hook to pass the previous recurring contribution ID (the recurring contribution before the renewal in case of payment plan payments) as a third parameter.

### DIFF
--- a/CRM/MembershipExtras/Form/PaymentPlanSettings.php
+++ b/CRM/MembershipExtras/Form/PaymentPlanSettings.php
@@ -27,10 +27,16 @@ class CRM_MembershipExtras_Form_PaymentPlanSettings extends CRM_Core_Form {
     $settingFields  = $this->getSettingFields();
 
     foreach ($settingFields  as $name => $field) {
-      if ($name == 'membershipextras_paymentplan_default_processor') {
-        $this->addDefaultProcessorField($field);
-      } else {
-        $this->addSettingField($field);
+      switch ($name) {
+        case 'membershipextras_paymentplan_default_processor':
+          $this->addDefaultProcessorField($field);
+          break;
+        case 'membershipextras_customgroups_to_exclude_for_autorenew':
+          $this->addCustomGroupsToExcludeField($field);
+          break;
+        default:
+          $this->addSettingField($field);
+          break;
       }
 
       $settingFieldNames[] = $name;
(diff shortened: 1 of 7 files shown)


CRM/MembershipExtras/Form/PaymentPlanSettings.php
```diff
@@ -69,6 +75,37 @@ private function addDefaultProcessorField($defaultProcessorField) {
...
+      'sequential' => 1,
+      'return' => ["id", "title"],
+      'extends' => ['IN' => ['Contribution', 'ContributionRecur']],
+      'options' => ['limit' => 0],
+    ]);
+
+    $customGroupsOptions = [];
+    if (!empty($customGroups['values'])) {
+      foreach ($customGroups['values'] as $customGroup) {
+        $customGroupsOptions[$customGroup['id']] = $customGroup['title'];
```
MiyaNoctem: In the far off chance someone adds  a custom group with the same title to both Contributions and Recurring Contributions,  it might make sense to include the extended entity's name here, like `$customGroupsOptions[$customGroup['id']] = $customGroup['extends'] . ': ' . $customGroup['title']`.

CRM/MembershipExtras/Service/MembershipInstallmentsHandler.php
```diff
@@ -158,7 +158,11 @@ private function recordMembershipContribution($contributionNumber) {
       ]);
     }
 
-    $this->copyContributionCustomFields($contribution->id);
+    CRM_MembershipExtras_Service_CustomFieldsCopier::copy(
+      $this->lastContribution['id'],
+      $contribution->id,
+      'Contribution'
+      );
```
MiyaNoctem: Indentation in this line seems off.

CRM/MembershipExtras/Form/PaymentPlanSettings.php
omarabuhussein: good point, thanks man!

PR #54, opened 2018-08-04
BASW-217: View a Recurring Contribution's Next Period Line Items

16kilobyte:
# Overiew
In the "Manage Installments" modal popup, we need to add a next period tab that should be shown if the recurring contribution has auto-renew enabled and it has at least one membership.
In the next period tab (default), each "offline_contribution_recur_line_item" with auto-renew = TRUE should be displayed. Each row should have the following information:
- Item - line item label
- Financial type - financial type of the line item
- Tax - the tax percentage of the financial type, show N/A if there is no tax account in the financial type
- Amount - line total of the line item
- Remove button - show as bin icon
- An "Add Membership" button and an "Add Other Amount" button should be shown under the last row

At the bottom of the tab, the following information should be shown:
- Untaxed Amount - total amount of the line_totals
- Tax - total amount of tax_amounts
- Total per installment - sum of a and b

# Before
Action did not exist

# After
Action was added only to offline manual payment recurring contributions.

### DIFF
--- a/membershipextras.civix.php
+++ b/membershipextras.civix.php
@@ -269,10 +269,10 @@ function _membershipextras_civix_civicrm_managed(&$entities) {
       if (empty($e['module'])) {
         $e['module'] = E::LONG_NAME;
       }
-      $entities[] = $e;
       if (empty($e['params']['version'])) {
         $e['params']['version'] = '3';
       }
+      $entities[] = $e;
     }
   }
 }


MiyaNoctem:
There seems to be something wrong here, as those commits by Omar shouldn't really be there. I think you may have based your branch off develop which I guess has those commits, while the feature branch doesn't. I'm going to pull in those commits before we can merge this one, but try to create the next branch off the feature branch nex time, so we don't have this problem.

PR #56, opened 2018-08-08
BASW-218: Remove Payment Plan Line Items

MiyaNoctem:
## Overview
As part of the management of line items associated to a payment plan (recurring contribution), we need to add a way to remove those line items and have that affect any payments that are still pending for the payment plan. To do this, the trash -bin icon on the line item list for the recurring contribution should be used to implement the removal.

1. Once the bin button is clicked, if the item is the last item in the current period tab, a notice with warning sign should appear with text "Cannot remove the last item in an order" and action should be cancelled.
2. Once the bin button is clicked, if the item is not the last item in the current period tab, a dialog with title "Remove LINE_ITEM_LABEL ?" should appear with the following components:
- Dialog body text: "LINE_ITEM_LABEL amount should be deducted from all remaining instalments after the end date. Please note the changes should take effect immediately after "Apply"."
- Adjust end date: checkbox option
- End date: datepicker that should only be shown if "Adjust end date" is checked, default to today's date
- Cancel button
- Apply button

3. If user clicks on "Apply":
- end_date in offline_contribution_recur_line_item for the corresponding item should be filled in with today's date. If "Adjust end date" is checked, the end_date should be set to the date user picked.
- auto_renew in offline_contribution_recur_line_item for the corresponding item should set to False.
- For any "Pending" instalment contribution with "Received Date" equals to or after the end_date, any line_item with same label as the recurring line item should be removed. Contribution amounts should be adjusted.
- If the line item is a membership, update the corresponding membership (linked to the recurring contribution) to have the new end_date and clear the contribution_recurr_id
- A notice should appear with text "LINE_ITEM_LABEL has been removed from the active order."
- A notice should appear with text "LINE_ITEM_LABEL should no longer be continued in the next period."
- The content of "Manage Instalments" modal should be reloaded.

## Before
The bin button was not functional.

## After
Implemented removal using a new form, which is called in a new dialog when the bin button is clicked. The form shows the required fields (the checkbox and the date picker), with the date picker disabled by default, only being enabled if the checkbox option is clicked.

The form is submitted when clicking on the *Apply* button. Post-processing of the form is used to:
- If line item is a membership, it cancels the membership.
- Cycle through pending contributions after selected end date to set corresponding line item amounts to zero, update contribution amounts by subtracting the line item's original amount and creating financial transaction records associated to the contributions to log the decrease of the contributions's amount.
- Update recurring line item to set its end-date and cancel auto-renew.

Once the form is processed, the list of line items is refreshed to show the changes made.

basw-218-after

### DIFF
--- a/templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
+++ b/templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
@@ -1,3 +1,45 @@
+<script>
+  var recurringContributionID = {$recurringContributionID};
+
+  {literal}
+  CRM.$(function () {
+    CRM.$('.remove-line-button').each(function () {
+      CRM.$(this).click(function () {
+        var itemID = CRM.$(this).attr('itemid');
+        showLineItemRemovalConfirmation(itemID);
+
+        return false;
+      });
+    });
+  });
+
+  function showLineItemRemovalConfirmation(lineItemID) {
+    CRM.api3('ContributionRecurLineItem', 'getcount', {
+      'contribution_recur_id': recurringContributionID,
+      'end_date': {'IS NULL': 1},
+    }).done(function (result) {
+      if (result.result < 2) {
+        CRM.alert("Cannot remove the last item in an order!", null, 'alert');
+
+        return;
+      }
+
+      var formUrl = CRM.url('civicrm/recurring-contribution/remove-lineitems', {
+        reset: 1,
+        contribution_recur_id: recurringContributionID,
+        line_item_id: lineItemID
+      });
+
+      CRM.loadForm(formUrl, {
+        dialog: {width: 480}
+      }).on('crmFormSuccess', function(event, data) {
+        CRM.refreshParent('#periodsContainer');
+      });
+    });
+  }
+  {/literal}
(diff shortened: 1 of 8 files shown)


templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
```diff
@@ -22,17 +64,18 @@
...
       <td>{$currentItem.end_date|date_format}</td>
-      <td><input type="checkbox" disabled{if $currentItem.auto_renew} checked{/if} /></td>
+      <td><input type="checkbox"
+                 disabled{if $currentItem.auto_renew} checked{/if} /></td>
       <td>{$currentItem.financial_type}</td>
       <td>{$currentItem.tax_amount|crmMoney}</td>
       <td>{$currentItem.line_total|crmMoney}</td>
       <td>
-        <a class="delete" href="">
+        <a class="remove-line-button" href="#" itemid="{$currentItem.id}">
```
vinuvarshith: Can you pls use 'data-itemid' for the attribute name instead of 'itemid' as this will be invalid HTML
Then use jquery .data() to get the attribute
Thanks

PR #57, opened 2018-08-08
 BASW-206: Fix: Change auto_renew Field Type To Boolean From Datetime 

16kilobyte:
# Overview
As per the ticket (BASW-206) overview, `auto_renew` should be boolean but was implemented as a `datetime` field.

# Before
`auto_renew` was `datetime`

# After
`auto_renew` is a boolean field and copied in the `LineItem` post hook as a `boolean` instead of a `string`


sql/auto_uninstall.sql
```diff
@@ -1,23 +1,38 @@
...
+-- Generated from drop.tpl
+-- DO NOT EDIT.  Generated by CRM_Core_CodeGen
+--
 -- /*******************************************************
--- * Delete External IDs value tables
+-- *
+-- * Clean up the exisiting tables
+-- *
 -- *******************************************************/
-DROP TABLE IF EXISTS `civicrm_value_contribution_ext_id`;
```
guanhuan: @16kilobyte should these lines to removed?

sql/auto_uninstall.sql
```diff
@@ -1,23 +1,38 @@
...
-DROP TABLE IF EXISTS `civicrm_value_contribution_recur_ext_id`;
-DROP TABLE IF EXISTS `civicrm_value_membership_ext_id`;
-DROP TABLE IF EXISTS `civicrm_value_line_item_ext_id`;
 
--- /*******************************************************
--- * Delete Relationships between recurring contributions
--- * and line items
--- *******************************************************/
-DELETE FROM civicrm_line_item WHERE `id` IN (
```
guanhuan: Also these

sql/auto_uninstall.sql
```diff
@@ -1,23 +1,38 @@
-SET FOREIGN_KEY_CHECKS=0;
-
-DELETE FROM civicrm_setting WHERE `name` LIKE 'membershipextras_paymentplan_%';
```
guanhuan: looks like the old uninstall script is overwritten.

sql/auto_install.sql
```diff
@@ -1,27 +1,101 @@
...
-    FOREIGN KEY (`contribution_recur_id`) REFERENCES `civicrm_contribution_recur`(`id`)
-    ON DELETE CASCADE,
-
-  CONSTRAINT FK_membershipextras_contribrecur_lineitem_line_item_id
-    FOREIGN KEY (`line_item_id`) REFERENCES `civicrm_line_item`(`id`)
-    ON DELETE CASCADE
-);
+
+
+     `id` int unsigned NOT NULL AUTO_INCREMENT  COMMENT 'Discount Item ID',
```
guanhuan: This block has some indentation and line break problems

sql/auto_install.sql
```diff
@@ -1,27 +1,101 @@
...
+ 
+    ,     UNIQUE INDEX `index_contribrecurid_lineitemid`(
+        contribution_recur_id
+      , line_item_id
+  )
+  
+,          CONSTRAINT FK_membershipextras_contribrecur_lineitem_contribution_recur_id FOREIGN KEY (`contribution_recur_id`) REFERENCES `civicrm_contribution_recur`(`id`) ON DELETE CASCADE,          CONSTRAINT FK_membershipextras_contribrecur_lineitem_line_item_id FOREIGN KEY (`line_item_id`) REFERENCES `civicrm_line_item`(`id`) ON DELETE CASCADE  
+)    ;
+
+
```
MiyaNoctem: @16kilobyte, normally we don't alter autogenerated files like this one, but this is actually an exception. We normally edit these to clean them up a bit (indentation, styling and the like). The reason is sometimes there is more code in here, other than the one generated by civix, and we need to keep track of all those DB changes needed on either install or uninstall. Also, you can remove all the comments at the start, kind of leaving it as it was.

BTW, according to our coding standard, all files should end in a blank line, so please remember to add this once you do the fix.

sql/auto_uninstall.sql
MiyaNoctem: This is an example of what I said in my previous comment. These lines can't be deleted from the uninstallation script, so you'll need to add them in again.

sql/auto_uninstall.sql
MiyaNoctem: They definitely should not be removed!

sql/auto_uninstall.sql
```diff
@@ -1,23 +1,38 @@
...
-DELETE FROM civicrm_line_item WHERE `id` IN (
-  SELECT membershipextras_contribrecur_lineitem.line_item_id
-  FROM membershipextras_contribrecur_lineitem
-);
+SET FOREIGN_KEY_CHECKS=0;
+
 DROP TABLE IF EXISTS `membershipextras_contribrecur_lineitem`;
 
-SET FOREIGN_KEY_CHECKS=1;
+SET FOREIGN_KEY_CHECKS=1;
```
MiyaNoctem: Blank line at EOF.

sql/auto_install.sql
```diff
@@ -1,27 +1,94 @@
...
+-- | License and the CiviCRM Licensing Exception along                  |
+-- | with this program; if not, contact CiviCRM LLC                     |
+-- | at info[AT]civicrm[DOT]org. If you have questions about the        |
+-- | GNU Affero General Public License or the licensing of CiviCRM,     |
+-- | see the CiviCRM license FAQ at http://civicrm.org/licensing        |
+-- +--------------------------------------------------------------------+
+--
+-- Generated from schema.tpl
+-- DO NOT EDIT.  Generated by CRM_Core_CodeGen
+--
```
MiyaNoctem: Please delete all these comments, from line 1 to line 27.

sql/auto_install.sql
```diff
@@ -1,27 +1,94 @@
...
+-- | You should have received a copy of the GNU Affero General Public   |
+-- | License and the CiviCRM Licensing Exception along                  |
+-- | with this program; if not, contact CiviCRM LLC                     |
+-- | at info[AT]civicrm[DOT]org. If you have questions about the        |
+-- | GNU Affero General Public License or the licensing of CiviCRM,     |
+-- | see the CiviCRM license FAQ at http://civicrm.org/licensing        |
+-- +--------------------------------------------------------------------+
+--
+-- Generated from drop.tpl
+-- DO NOT EDIT.  Generated by CRM_Core_CodeGen
```
MiyaNoctem: All these comments should also be deleted. Seems they're actually duplicated.

sql/auto_install.sql
16kilobyte: Done

sql/auto_install.sql
16kilobyte: Done

PR #60, opened 2018-08-17
BASW-220: Add New Membership to Current Period

MiyaNoctem:
## Overiew
We need to implement a way to add new memberships to a recurring contribution. For this, we should use the 'Add Membership' button available on the 'View/Manage Line Items' page of a recurring contribution. The idea is when the button is clicked, a new line item with empty values should be added to the table, allowing the user to input appropriate values an click on a check icon at the end of the row once he wants to apply the change.

Clicking the check icon should ask for confirmation, allowing the user to finally decide if he wants to change the amount for the first pending contribution.

Once user confirms, the new line item should be added to the recurring contribution, and new line items should be created for each pending contribution after the new membership's start date.

## Before
Add Membership button did nothing.

## After
Added front end logic to:
- Add new row with inputs when clicking on 'Add Membership' button.
- Handle events when clicking check, and cross icons, to either show confirmation form or hide new line respectively.
- Disable all orher clickable elements on the form while the new row is being shown.
- Make new row flash if anythin else is clicked, other than elements in the new row.

All the JS logic was refactored into a class and moved to a separate file, to declutter the template.

Once user confirms addition, confirmation form back-end post-processing deals with:
- Creates or updates membership with given dates.
- Creates new line item and, relating it to the current recurring contribution.
- Goes through all pending payments after selected start date to add or update corresponding line items.
- Creates financial transaction with its financial line items to record the change in the contribution amount.

### Initial Conditions
The initial state will start with a contact having a single membership associated to a 12-month payment plan and a donation.

- General 1yr membership: $100
- Donation II: $1200
- 20% tax
- **Total: $1560**
- **Amount per Installation: $130**

basw-220-initial-membership

### Contact Doesn't Have Membership Type
A new Gold membership will be added, for a total of $12000 + 20% tax.

basw-220-no-previous-membership-type

### New Membership With Different First Amount
A new Gold membership will be added, for a total of $12000 + 20% tax, but the first contribution will only add $10 20% tax.

basw-220-diferent-first-amount


guanhuan:
@MiyaNoctem can we also include the followings examples?
- an example where the first instalment amount is different
- an example where the contact did not have the selected membership type before the line is added

You can split these in different gifs for different scenarios.

For each scenario, can we also include the following demo?
- viewing the instalments after a line is added
- viewing lines of one of the instalments that were affected
- viewing the affected membership before and after

guanhuan:
Hi @MiyaNoctem, this looks great!

Two minor points:

1. (new change) After watching your recording, I feel the current wording is not informative enough for the user. I have just updated the modal wording to be:

> Please note the new line will be added to all pending instalments starting from START_DATE immediately after clicking "Apply".

Could you update the PR to reflect that?

2. Some minor UI adjustments.

- Could you make sure the elements in the modal follows CiviCRM markup (you can chat with Alessandro/ Swastik about this)? 
- Currently the elements look a bit jumbled, maybe it's the width of the text input etc? 
- Could you set a min width/ height for the modal?

MiyaNoctem:
@guanhuan all markup uses normal CiviCRM markup. I did reduce the size of the only input that would allow it, namely the membership's amount. I don't think I could/should reduce the size of date fields, due to the calendar icon they have, nor the select of the membership type, which automatically adjusts to the length of the options.

PR #61, opened 2018-08-24
BASW-223: Add Other Amount Line Item To Multi Line Order

16kilobyte:
## Overview
1. In the "View/Modify Future Instalments" next period tab, user can use "Add Membership" and "Add Other Amount" buttons to add line items to the current order.
2. Once the "Add Other Amount" is clicked, a new row should appear in next period tab with the following information:
- Item - required - a text box
- Financial type - select list with all available financial types
- Tax - text showing the tax percentage of the selected financial type, show N/A if there is no tax account in the selected financial type
- Amount - required - a money input box
- cross icon - cancel action
- check icon - confirm action
3. When a row is added but not yet confirmed/ cancelled, any parts of the modal popup other then the row itself should be made unclickable. If the user tries to click on any other parts, the row should flash to remind the user an item is yet to be completed.
4. If user clicked on cross icon, the row should be cancelled and no change is made.
5. When user clicked on the check icon, a dialog with title "Add LINE_ITEM_LABEL ?" should appear with the following components:
- dialog body text - "Please note the changes should take effect immediately after "Apply"."
- Cancel button
- Apply button
6. If user click on "Apply" button:
- a line item with empty contribution_id should be created according to the information user inputted. Note that the tax_amount will calculated by multiplying amount with tax rate.
- a offline_contribution_recur_line_item should be created linking the line item and the recurring contribution. It's start date and end date should be empty.
- a notice should appear with text "LINE_ITEM_LABEL will now be continued in the next period."
- the content of "Manage Instalments" modal should be reloaded.
7. If user clicked on "Cancel", the action should be cancelled and no change is made.
8. User can also check "Renew automatically" field for any of other amount lines in "Current period" tab to perform the same action. Point 5-7 also apply here.

## Before
The bin button was not functional.

## After
Action has been implemented.

basw-223-after

### DIFF
--- a/templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
+++ b/templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
@@ -1,21 +1,138 @@
 <script>
 var recurringContributionID = {$recurringContributionID};
+var financialTypes = JSON.parse('{$financialTypes|@json_encode}');
+var disableClicks = false;
 
 {literal}
   CRM.$(function () {
+
     CRM.$('.remove-next-period-line-button').each(function () {
       CRM.$(this).click(function (e) {
         e.preventDefault();
+        
+        if(!disableClicks) {
+          var itemData = CRM.$(this).closest('tr').data('item-data');
+          showNextPeriodLineItemRemovalConfirmation(itemData);
 
-        var itemData = CRM.$(this).closest('tr').data('item-data');
-        showNextPeriodLineItemRemovalConfirmation(itemData);
-
-        CRM.$('#periodsContainer').on('crmLoad', function(event, data) {
-          CRM.$('#tab_next a').click();
-        });
+          CRM.$('#periodsContainer').on('crmLoad', function(event, data) {
+            CRM.$('#tab_next a').click();
+          });
+        }
       });
     });
+
+    CRM.$('#next_buttons #addOtherAmount').on('click', function(e) {
+      e.preventDefault();
+
+      if (!disableClicks) {
+        CRM.$('#addLineItemRow').show();
+        CRM.$('#periodsContainer').find('*').not(CRM.$('#addLineItemRow').find('*')).css('cursor', 'not-allowed');
+        CRM.$('#addLineItemRow').css('cursor', 'auto');
+        CRM.$('#addLineItemRow').css('opacity', 1);
+
+        disableClicks = true;
+      }
+    });
+
+    CRM.$('.cancel-add-next-period-line-button').on('click', function(e) {
+      e.preventDefault();
+
+      CRM.$('#addLineItemRow').hide();
+      CRM.$('#periodsContainer').find('*').css('cursor', 'auto');
+    });
+
+    CRM.$('#financialType').on('change', function() {
+      var selectedId = CRM.$(this).val();
+      var financialType = getFinancialType(selectedId);
+
+      if (!financialType) {
+        throw new Error('Invalid financial type id passed');
+      }
+      
+      CRM.$('#financialTypeTaxRate').text(financialType.tax_rate || 'N/A');
+    });
+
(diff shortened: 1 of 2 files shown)


guanhuan:
Thanks @16kilobyte.

Adding the other amount to the next period shouldn't add it to the current period. This doesn't look quite right from the gif.

Please also note the ticket title is "Add an other amount line to a standard multi-line membership order - next period only" but the PR seems to state this is for general addition.

Could you clarify?

MiyaNoctem:
On @guanhuan 's note, currently only line items with end date are being filtered out from current period tab. I think we should filter out line items without a start date, that'd take care of line items added forn next period only.

16kilobyte:
@MiyaNoctem, clear?

16kilobyte:
@guanhuan,
Current line items are not filtered so all line items are listed.
I have updated the PR's title.

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -1,22 +1,92 @@
...
+        try {
+          amount = parseInt(amount);
+        } catch(error) {
+          CRM.alert('Amount you entered is not valid', null, 'error');
+          return;
+        }
+      }
+
+      showAddOtherAmountConfirmation(label, amount, financial_type_id)
+
```
MiyaNoctem: You can delete this blank line.

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -70,12 +140,63 @@ var recurringContributionID = {$recurringContributionID};
...
+          qty: 1.0,
+          unit_price: amount,
+          line_total: amount,
+          tax_amount: taxAmount,
+          financial_type_id: financial_type_id,
+          entity_table: 'civicrm_contribution_recur',
+        }).done(function(lineItemResult) {
+          if (lineItemResult.is_error) {
+            CRM.alert(lineItemResult.error_message, null, 'error');
+            return;
```
MiyaNoctem: Blank line before return.

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -70,12 +140,63 @@ var recurringContributionID = {$recurringContributionID};
...
+
+          var createdLineItemId = lineItemResult.id;
+          CRM.api3('ContributionRecurLineItem', 'create', {
+            contribution_recur_id: recurringContributionID,
+            line_item_id: createdLineItemId,
+            auto_renew: true,
+          }).done(function(result) {
+            if (result.is_error) {
+              CRM.alert(result.error_message, null, 'error');
+              return;
```
MiyaNoctem: Blank line before return.

guanhuan:
Got it @MiyaNoctem @16kilobyte. On the current period tab, let's filter out the ones that don't have start date too.

@16kilobyte Otherwise this looks very good! I do want to re-prioritise the tab after reload issue i.e. when actions are taken on next period tab, the modal should land on next period tab after modal reloads. Could you target at resolving that before the next ticket? I think usability-wise this is very important to sort out?

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -1,22 +1,92 @@
...
+        CRM.alert('Item label is required', null, 'error');
+        return;
+      }
+
+      if (!amount.length) {
+        CRM.alert('Item amount is required', null, 'error');
+        return;
+      } else {
+        try {
+          amount = parseInt(amount);
```
MiyaNoctem: Amounts can be floating point values, so you should probably use `parseFloat` instead.

MiyaNoctem: Maybe `isNan()` would be better?

PR #63, opened 2018-08-29
BASW-223: Fix Modal Reload Auto Switch To Next Period Tab

16kilobyte:
## Overview
As per the ticket (BASW-223) overview, after modal pop reloads when a new other amount is added, it should automatically switch to the next period tab

## Before
After modal popup reloads, it does not automatically switch to the next period tab.

## After
Modal popup switches to next period tab when other amount is added

basw-219-after

### DIFF
--- a/templates/CRM/MembershipExtras/Page/EditContributionRecurLineItems.tpl
+++ b/templates/CRM/MembershipExtras/Page/EditContributionRecurLineItems.tpl
@@ -1,6 +1,7 @@
 <div id="periodsContainer" class="ui-tabs ui-widget ui-widget-content ui-corner-all">
   <script type="text/javascript">
-    var selectedTab  = 'current';
+    var selectedTab  = window.CompucorpMembershipExtras_selectedTab || 'current';
+    window.CompucorpMembershipExtras_selectedTab = 'current';
     {literal}
     CRM.$(function($) {
       var tabIndex = $('#tab_' + selectedTab).prevAll().length;
@@ -11,7 +12,7 @@
   </script>
 
   <ul class="ui-tabs-nav ui-corner-all ui-helper-reset ui-helper-clearfix ui-widget-header">
-    <li id="tab_current" class="crm-tab-button ui-corner-all ui-tabs-tab ui-corner-top ui-state-default ui-tab ui-tabs-active ui-state-active">
+    <li id="tab_current" class="crm-tab-button ui-corner-all ui-tabs-tab ui-corner-top ui-state-default ui-tab">
       <a href="#current-subtab" title="{ts}Contributions{/ts}">
         {ts}Current Period{/ts}
       </a>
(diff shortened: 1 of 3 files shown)


templates/CRM/MembershipExtras/Page/EditContributionRecurLineItems.tpl
```diff
@@ -1,6 +1,7 @@
 <div id="periodsContainer" class="ui-tabs ui-widget ui-widget-content ui-corner-all">
   <script type="text/javascript">
-    var selectedTab  = 'current';
+    var selectedTab  = window.CompucorpMembershipExtras_selectedTab || 'current';
+    window.CompucorpMembershipExtras_selectedTab = 'current';
```
MiyaNoctem: I don't think we need the PR for 219 if we're having this in this PR.

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -1,22 +1,83 @@
...
       CRM.$(this).click(function (e) {
         e.preventDefault();
-
         var itemData = CRM.$(this).closest('tr').data('item-data');
         showNextPeriodLineItemRemovalConfirmation(itemData);
 
-        CRM.$('#periodsContainer').on('crmLoad', function(event, data) {
-          CRM.$('#tab_next a').click();
-        });
+        window.CompucorpMembershipExtras_selectedTab = 'next';
```
MiyaNoctem: I don't think this should happen here only if the remove icon is clicked. I think we should add this logic to actually clicking the tabs instead and move it to EditContributionRecurLineItems.tpl.

Eg.
```javascript
var selectedTab = window.CompucorpMembershipExtras_selectedTab || 'current';

CRM.$(function($) {
  CRM.$('#tab_current').click(function () {
    window.CompucorpMembershipExtras_selectedTab = 'current';
  });

  CRM.$('#tab_next').click(function () {
    window.CompucorpMembershipExtras_selectedTab = 'next';
  });

  var tabIndex = $('#tab_' + selectedTab).prevAll().length;
  $("#periodsContainer").tabs({active: tabIndex});
  $(".crm-tab-button").addClass("ui-corner-bottom");
});
```

16kilobyte: The problem with this approach is that whatever tab the user closed the modal popup on will be reopened if the user reopens the popup without refreshing the page. Is that a desired effect?

MiyaNoctem: Hmm, good point. Maybe if I open the same recurring contribution it could make sense, but it might be awkward if I open another recurring contribution. What if we add an attribute to the dialog's markup to store the value instead? Something like:

```javascript
var selectedTab = CRM.$(#periodsContainer).closest('.ui-dialog-content').data('selectedTab') || 'current';
...
CRM.$(#periodsContainer).closest('.ui-dialog-content').data('selectedTab', 'current');
...
CRM.$(#periodsContainer).closest('.ui-dialog-content').data('selectedTab', 'next');
```

That way, every time the dialog is opened, the value would be empty.

PR #64, opened 2018-08-31
BASW-246: Update And Refactor

16kilobyte:
## Overview
We are required to:
1. Update the table name across the board to be `membershipextras_subscription_line`
2. Specify `FALSE` as the default value for `auto_renew column` in the table
3. Add an `is_removed` boolean column and specify `FALSE` as the default value

## Before
Table was called "membershipextras_contribrecur_lineitem", `auto_renew` did not default to `FALSE` but to `NULL` and `is_removed` field was not existing.

## After
Table is now called "membershipextras_subscription_line", `auto_renew` defaults to `FALSE` and `is_removed` field has been added.


membershipextras.civix.php
```diff
@@ -269,10 +269,10 @@ function _membershipextras_civix_civicrm_managed(&$entities) {
       if (empty($e['module'])) {
         $e['module'] = E::LONG_NAME;
       }
+      $entities[] = $e;
```
MiyaNoctem: I think this is a bug that's fixed on latest versions of civix. I believe this line should be removed from here and added to line 275, as it was before.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -179,4 +179,15 @@ public function upgrade_0001() {
     return TRUE;
   }
 
+  /**
+   * Renames membershipextras_contribution_recur_line_item table to membershipextras_subscription_line;.
+   *
+   * @return bool
+   */
+  public function upgrade_0002() {
```
MiyaNoctem: There's no need for this upgrade, per Guanhuan's instructions. The idea is to have only one upgrader for the whole of phase 3 (that would be upgrade_0001), and since we've done the changes on the auto_install.sql file, it is enough. This will require you and me to uninstall the extension and re-install it to keep on working, but that's the idea.

sql/rename_table_and_add_field.sql
```diff
@@ -0,0 +1,6 @@
+RENAME TABLE `membershipextras_contribrecur_lineitem` TO `membershipextras_subscription_line`;
```
MiyaNoctem: As explained before, this file isn't needed, so please remove it.

CRM/MembershipExtras/Upgrader.php
16kilobyte: @guanhuan mentioned that people are already using the extension

PR #65, opened 2018-09-03
BASW-221: Add Other Contribution Line Item

MiyaNoctem:
## Overview
We need to add new line items to a recurring contribution, different to a membership. For this, we should use the 'Add Other Amount' button available on the 'View/Manage Line Items' page of a recurring contribution. The idea is when the button is clicked, a new line item with empty values should be added to the table, allowing the user to input appropriate values and click on a check icon at the end of the row once he wants to apply the change.

Clicking the check icon should ask for confirmation, allowing the user to finally decide if he wants to change the amount for the first pending contribution.

Once user confirms, the new line item should be added to the recurring contribution, and new line items should be created for each pending contribution after the start date selected by the user.

## Before 
Clicking on 'Add Other Amount Button' did nothing.

## After
Added front end logic to:
- Add new row with inputs when clicking on 'Add Other Amount' button.
- Handle events when clicking check, and cross icons, to either show confirmation form or hide new line respectively.
- Disable all other clickable elements on the form while the new row is being shown.
- Make new row flash if anything else is clicked, other than elements in the new row.

Once user confirms addition, confirmation form back-end post-processing deals with:
- Creates new line item and, relating it to the current recurring contribution.
- Goes through all pending payments after selected start date to add or update corresponding line items.
- Creates financial transaction with its financial line items to record the change in the contribution amount.

basw-221-after

## Notes
Code done for BASW-220 was reused by refactoring code common to both line item addition forms into a parent abstract class.


CRM/MembershipExtras/Form/RecurringContribution/AddDonationLineItem.php
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  /**
+   * Creates new line item associated to the recurring contribution.
+   *
+   * @return array
+   */
+  protected function createRecurringLineItem() {
+    $taxRate = $this->getTaxRateForFinancialType($this->lineItemParams['financial_type_id']);
+    $taxAmount = round($this->lineItemParams['amount'] * $taxRate / 100, 2);
```
omarabuhussein: can you check all the places that use round($amount/100, 2) formula and move it to a new class (that could probably called AmountCalculator or whatever) instead of repeating the same line in the extension codebase multiple times.

css/style.css
```diff
@@ -14,10 +14,11 @@ tr.disabled-row {
...
-#newline_membership_type.required {
+#newline_membership_type.required,
+#newline_donation_financial_type_id.required {
   border: 2px solid #900 !important;
 }
 
-#newline_amount {
+#newline_amount, #newline_item, #newline_donation_amount, #newline_donation_item {
   width: 50%;
 }
```
omarabuhussein: new line here needed

guanhuan:
It looks like the recurring contribution amount is not updated after added the line. @MiyaNoctem Could you check this and make sure the recurring contribution amount is updated every time something is added/ removed from the current period.

PR #70, opened 2018-09-17
BASW-222: Add Membership Line to Standard Multi-line Membership Order - Next Period

16kilobyte:
# Overiew
In the "View/Modify Future Instalments" next period tab, user can use "Add Membership" and "Add Other Amount" buttons to add line items to the current order.
Once the "Add Membership" is clicked, a new row should appear in next period tab with the following information:
- Item - required - a dropdown contains all membership types which are not belong to any of the membership organisation for the existing lines
- Financial type - text showing the financial type of the selected membership
- Tax - text showing the tax percentage of the financial type, show N/A if there is no tax account in the financial type
- Amount - required - a money input box defaulted to the minimum_fee of the membership type divided by number_of_instalments.
- cross icon - cancel action
- check icon - confirm action

When a row is added but not yet confirmed/ cancelled, any parts of the modal popup other then the row itself should be made unclickable. If the user tries to click on any other parts, the row should flash to remind the user an item is yet to be completed.

If user clicked on cross icon, the row should be cancelled and no change is made.

When user clicked on the check icon, a dialog with title "Add LINE_ITEM_LABEL ?" should appear with the following components:
- dialog body text - "Please note the changes should take effect immediately after "Apply"."
- Cancel button
- Apply button

If user click on "Apply" button:
- a line item with empty contribution_id should be created according to the information user inputted. Note that the tax_amount will calculated by multiplying amount with tax rate.
- a *membershipextras_subscription_line* should be created linking the line item and the recurring contribution. It's start date and end date should be empty.
- a notice should appear with text "LINE_ITEM_LABEL will now be continued in the next period."
- the content of "Manage Instalments" modal should be reloaded.

If user clicked on "Cancel", the action should be cancelled and no change is made.

User can also check "Renew automatically" field for any of membership lines in "Current period" tab to perform the same action. A notice with warning sign will appear with text "This membership type is already enrolled in next period." when there is another *membershipextras_subscription_line*:
- linked to a line item with same membership type and *membershipextras_subscription_line*
- end date is empty
- auto-renew is TRUE

If user checks the "Renew automatically" field for any of membership lines in "Current period" tab and the scenario in point 8 does not apply, point 5-7 should apply here.

# Before
Action did not exist

# After
Action was added only to offline manual payment recurring contributions.


membershipextras.php
```diff
@@ -349,5 +349,12 @@ function membershipextras_civicrm_pageRun($page) {
       1,
       'page-header'
     );
+
+    CRM_Core_Resources::singleton()->addScriptFile(
```
MiyaNoctem: It is customary to chain these calls like this:

```php
CRM_Core_Resources::singleton()->addScriptFile(
  CRM_MembershipExtras_ExtensionUtil::LONG_NAME,
  'js/CurrentPeriodLineItemHandler.js',
  1,
  'page-header'
)->addScriptFile(
CRM_MembershipExtras_ExtensionUtil::LONG_NAME,
  'js/NextPeriodLineItemHandler.js',
  1,
  'page-header'
);
```

16kilobyte:
@guanhuan I have added a second screenshot

PR #72, opened 2018-09-18
BASW-217: Fix View Modify Future Installments Action - Next Period

16kilobyte:
## Overview
1. Line Items should display with percentage symbol next.
2. If the number of installments of the payment plan is empty or 0, `membershipextras_subscription_line` with an end date should be excluded from the view.

## Before
1. Percentage sign does not show next to tax rate.
2. Line items show regardless of if the payment plan has 0 installments and if the `membershipextras_subscription_line`  has an `end_date`.

## After
1. Percentage sign shows next to the tax rate now.
2. `membershipextras_subscription_line` are excluded when the payment plan has 0 installments and the `membershipextras_subscription_line`  has an `end_date`.

basw-217-fix-after

### DIFF
--- a/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
+++ b/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
@@ -145,11 +145,11 @@ class CRM_MembershipExtras_Page_EditContributionRecurLineItems extends CRM_Core_
     $this->assign('membershipTypes', $this->getAvailableMembershipTypes($currentPeriodLineItems));
     $this->assign('lineItems', $currentPeriodLineItems);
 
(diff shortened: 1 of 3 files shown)


guanhuan:
@16kilobyte I don't think this is quite correct, at least the description in "After - point 2".

The requirement is "If the number of installments of the payment plan is empty or 0, membershipextras_subscription_line with an end date should also be excluded from the view."

This is broken down to:
if "the number of installments of the payment plan is empty or 0" & "membershipextras_subscription_line with an end date", membershipextras_subscription_line should not show.

1. Please clarify above or make changes if needed.
2. Also please confirm this is for next period only in the description.

16kilobyte:
@guanhuan I have updated the description now.

PR #76, opened 2018-09-21
BASW-264: Payment Plan Period Links

16kilobyte:
## Overview
1. Creating a custom fieldset "Related Payment Plan Periods" for recurring contribution entity. It should contain two read-only integer fields:
  - Previous Payment Plan Period
  - Next Payment Plan Period
2. We need to also add this to an upgrader which is only run when upgrading process.

## Before
Feature was not present.

## After
Feature has been implemented. 

Payment Plan Period Links

### DIFF
--- a/CRM/MembershipExtras/Upgrader.php
+++ b/CRM/MembershipExtras/Upgrader.php
@@ -180,7 +180,7 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
       'collapse_display' => 1,
       'weight' => 10,
       'is_active' => 0,
-      'table_name' => 'civicrm_value_related_payme_21',
+      'table_name' => 'civicrm_value_payment_plan_periods',
       'is_multiple' => 0,
       'collapse_adv_display' => 0,
       'is_reserved' => 0,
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Upgrader.php
```diff
@@ -168,13 +168,65 @@ private function removeCustomExternalIDs() {
...
+  private function createPeriodLinkCustomFields() {
+    civicrm_api3('CustomGroup', 'create', [
+      'name' => 'related_payment_plan_periods',
+      'title' => E::ts('Related Payment Plan Periods'),
+      'extends' => 'ContributionRecur',
+      'style' => 'Inline',
+      'collapse_display' => 1,
+      'weight' => 10,
+      'is_active' => 0,
+      'table_name' => 'civicrm_value_related_payme_21',
```
MiyaNoctem: I'm not sure why you chose this table name, I guess you created the custom group using either the API or CiviCRM and this was the name generated the system. Thing is if we define the name of the table here and the xml, we normally try to give it a more meaningful name, eg. `civicrm_value_plan_period_linking` or something like that.

After you change the table's name, remember to update the xml also!

xml/customFields_install.xml
```diff
@@ -46,6 +46,23 @@
...
+            <name>related_payment_plan_periods</name>
+            <title>Related Payment Plan Periods</title>
+            <extends>ContributionRecur</extends>
+            <style>Inline</style>
+            <collapse_display>1</collapse_display>
+            <help_pre></help_pre>
+            <help_post></help_post>
+            <weight>10</weight>
+            <is_active>1</is_active>
+            <table_name>civicrm_value_related_payme_21</table_name>
```
MiyaNoctem: Remember to change the table's name here too, as per the latter observation.

PR #77, opened 2018-09-25
BASW-256: Set Subscription Line End Date

16kilobyte:
## Overview
When a payment plan is marked as completed, all subscription lines that meet the below conditions should have their end date set to the payment plan end date:
1. start date is not empty
2. end date is empty

## Before
Action did not exist

## After
When a payment plan's status changes to `Completed`, all subscription lines matching the criteria in the overview above are have their `end_date` set to the payment plan's `end_date`

### DIFF
--- a/CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
+++ b/CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
@@ -52,6 +52,7 @@ class CRM_MembershipExtras_Hook_Post_EntityFinancialTrxn {
           }
         }
       }
+
       civicrm_api3('ContributionRecur', 'create', [
         'id' => $this->recurContribution['id'],
         'contribution_status_id' => $newStatus,
@@ -59,6 +60,11 @@ class CRM_MembershipExtras_Hook_Post_EntityFinancialTrxn {
     }
   }
 
+  /**
+   * Returns LineItems associated to a recurring contribution
+   * 
+   * @return array
+   */
   private function getSubscriptionLines() {
     $lines = civicrm_api3('ContributionRecurLineItem', 'get', [
       'sequential' => 1,
--- a/CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
+++ b/CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
@@ -41,11 +41,31 @@ class CRM_MembershipExtras_Hook_Post_EntityFinancialTrxn {
 
     $newStatus = $this->generatePaymentPlanNewStatus();
     if ($newStatus !== NULL) {
+      if ($newStatus === 'Completed') {
+        $subscriptionLines = $this->getSubscriptionLines();
+        foreach($subscriptionLines as $line) {
+          if (!empty($line['start_date']) && empty($line['end_date'])) {
+            civicrm_api3('ContributionRecurLineItem', 'create', [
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
```diff
@@ -41,13 +41,39 @@ public function updatePaymentPlanStatus() {
 
     $newStatus = $this->generatePaymentPlanNewStatus();
     if ($newStatus !== NULL) {
+      if ($newStatus === 'Completed') {
```
omarabuhussein: you can extract the code here to a new private method to make it more clear and to avoid all of this code nesting : 

```
   if ($newStatus !== NULL) {
      civicrm_api3('ContributionRecur', 'create', [
        'id' => $this->recurContribution['id'],
        'contribution_status_id' => $newStatus,
      ]);
    }

    if ($newStatus === 'Completed') {
       $this->updateSubscriptionLinesEndDate();
     }


==========================================

    private function updateSubscriptionLinesEndDate() {
        $subscriptionLines = $this->getSubscriptionLines();
        foreach($subscriptionLines as $line) {
          if (!empty($line['start_date']) && empty($line['end_date'])) {
            civicrm_api3('ContributionRecurLineItem', 'create', [
              'id' => $line['id'],
              'end_date' => $this->recurContribution['end_date'],
            ]);
          }
         }
       }
```

CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
```diff
@@ -41,13 +41,39 @@ public function updatePaymentPlanStatus() {
...
 
+  /**
+   * Returns LineItems associated to a recurring contribution
+   * 
+   * @return array
+   */
+  private function getSubscriptionLines() {
+    $lines = civicrm_api3('ContributionRecurLineItem', 'get', [
+      'sequential' => 1,
+      'contribution' => $this->recurContribution['id'],
```
omarabuhussein: civicrm API has a hardcoded limit to get only 25 result from any API call, so for example if there was 28 subscritopn lines then only 25 ones will be fetched from this API call, you can alter the API to fetch all the results by adding : 

```
'options' => ['limit' => 0],
```

like this : 

```
    $lines = civicrm_api3('ContributionRecurLineItem', 'get', [
      'sequential' => 1,
      'contribution' => $this->recurContribution['id'],
      'options' => ['limit' => 0],
    ]);
```

CRM/MembershipExtras/Hook/Post/EntityFinancialTrxn.php
```diff
@@ -41,13 +41,39 @@ public function updatePaymentPlanStatus() {
...
+    ]);
+
+    return $lines;
```
omarabuhussein: have you tested this code ? because it would not work like this, civicrm API return the API rows under the 'values' array key. so you need to adjust this method as : 

```
   $subscriptionLinesResponse= civicrm_api3('ContributionRecurLineItem', 'get', [
      'sequential' => 1,
      'contribution' => $this->recurContribution['id'],
      'options' => ['limit' => 0],
    ]);

    $subscriptionLines = [];
    if (!empty($subscriptionLinesResponse['values'])){ 
      $subscriptionLines = $subscriptionLinesResponse['values'];
    }

    return  $subscriptionLines;
```

PR #80, opened 2018-10-03
BASW-262: Update Renewal with No Instalments to Use Subscription Line Items

MiyaNoctem:
## Overview

When "Renew offline auto-renewal memberships" scheduled job is run, any payment plans that meet criteria to auto-renew and have a number of instalments less than or equal to 1, should use line items on membershipextras_subscription_line to determine the line items for the new contribution.

For each membershipextras_subscription_line which is auto-renew and not is_removed in the recurring contribution, the system should:
- Make a copy of the linked line item and link it to the recurring contribution
- Make a copy of the membershipextras_subscription_line, linking it to the recurring contribution and making the start date of the copied membershipextras_subscription_line be the received date of the last instalment plus one interval. the end date should be empty.
- The end date of the original membershipextras_subscription_line should be 1 day earlier than the new membershipextras_subscription_line's start date.

The last contribution should be used as a template to create a new pending contribution under the same recurring contribution and should contain a copy of the auto-renew line items linked to the recurring contribution. The amount fields should reflect the actual amount based on the line items, both for the new contribution and the recurring contribution.

For each membership line item, if the user already have a membership with same type, we should extend that membership for a term.  If the user does not have any membership with same type, a new membership needs to be created with start date of the recurring contribution start date.

## Before
Line items to be included in new contribution was determined by line items set on the last contribution associated to the recurring contribution.

## After
Fixed some bugs that were causing problems on creation of payment plans with only one instalment, where the recurring line items were not being copied and associated to the recurring contribution.

Implemented logic to duplicate recurring line items that are set to renew, are not removed and do not have end dates, to be used for the new contribution, setting the end dates of the original line items.

Updated logic so the list of line items to be used on the new contribution on renewal is determined by those line items set to renew, are not removed and don't have an end date for the payment plan.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal.php
@@ -360,10 +360,11 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal {
       'contribution_recur_id' => $recurringContributionID,
       'auto_renew' => 1,
       'is_removed' => 0,
+      'end_date' => ['IS NULL' => 1],
       'api.LineItem.getsingle' => [
         'id' => '$value.line_item_id',
         'entity_table' => ['IS NOT NULL' => 1],
-        'entity_id' => ['IS NOT NULL' => 1]
+        'entity_id' => ['IS NOT NULL' => 1],
       ],
     ]);
 
@@ -596,6 +597,7 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal {
       'sequential' => 1,
       'contribution_recur_id' => $recorringContributionID,
       'start_date' => ['IS NOT NULL' => 1],
+      'end_date' => ['IS NULL' => 1],
       'is_removed' => 0,
       'api.LineItem.getsingle' => [
         'id' => '$value.line_item_id',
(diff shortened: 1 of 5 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -565,11 +565,55 @@ private function createRecurringContribution() {
...
+
+  /**
+   * Calculates amount for current recurring contribution from related line
+   * items.
+   */
+  private function calculateRecurringContributionTotalAmount($recorringContributionID) {
+    $totalAmount = 0;
+
+    $result = civicrm_api3('ContributionRecurLineItem', 'get', [
+      'sequential' => 1,
```
omarabuhussein: you need to use limit => 0 to get all the records.

CRM/MembershipExtras/Job/OfflineAutoRenewal.php
```diff
@@ -710,6 +713,48 @@ private function renewNoInstallmentsPaymentPlan() {
...
+  /**
+   * Creates duplicates of line items associated to recurring contribution that
+   * are set to auto-renew, don't have an end date and are not removed. It sets
+   * end date for the old line items.
+   *
+   * @param $recurringContributionID
+   */
+  private function duplicateRecurringLineItems($recurringContributionID) {
+    $lineItems = civicrm_api3('ContributionRecurLineItem', 'get', [
+      'sequential' => 1,
```
omarabuhussein: same here , you need to use limit => 0 to get all the records is needed

PR #82, opened 2018-10-08
BASW-259: Upgrader

16kilobyte:
### Part 1
1. As a part of phase 3 upgrader, the extension should find all pre-existing offline payment plans and populate the "Offline Recurring Contribution Line Item" information for them.
2. For each pre-existing recurring contribution using an offline payment processor (payment processor with Payment Manual class):
    * create a copy of each line item of the last installment contribution
    * the contribution_id of the copied line items should NULL
    * a record should be created in "membershipextras_subscription_line" table for each of the copied line item:
        * contribution_recur_id - the id of the recurring contribution
        * line_item_id - the id of the copied line item
        * start_date - start date of the recurring contribution
        * end_date - end date of the recurring contribution. If empty, leave empty.
        * auto_renew - true if the recurring contribution auto-renews, otherwise false
        * is_removed - False

### Part 2 (Needs to be run also during installation)
1. As a part of phase 3 upgrader, the extension should find all pre-existing payment plans that meet the criteria below and populate the "Related Payment Plan Periods" for them.
    * payment plan that uses an offline payment processor (payment processor with Payment Manual class)
    * payment plan whose number of installment is larger than 0
    * payment plan whose end date is not empty and is more a month earlier than today
    * payment plans whose auto-renew is set to True
2. the contact has more than one payment plans that meet criteria a, b and c
3. Find all memberships that are linked to payment plan that meets the criteria. Fill in the  "Previous Payment Plan Period" and "Next Payment Plan Period" with number 0.
4. This needs to be part of the installer too.

## Before
Action was not present.

## After
Action has been implemented.

### DIFF
--- a/CRM/MembershipExtras/Upgrader.php
+++ b/CRM/MembershipExtras/Upgrader.php
@@ -15,6 +15,7 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
     $this->createPaymentProcessor();
     $this->createLineItemExternalIDCustomField();
     $this->executeSqlFile('sql/set_unique_external_ids.sql');
+    $this->createOfflineSubLineItemsForPaymentPlans();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Upgrader.php
```diff
@@ -220,6 +221,162 @@ private function createPeriodLinkCustomFields() {
...
+    $lineItems = [];
+    if ($result['count'] > 0) {
+      $lineItems = $result['values'];
+    }
+
+    return $lineItems;
+  }
+
+  /**
+   *
```
MiyaNoctem: Missed the methods doc-block.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -220,6 +221,162 @@ private function createPeriodLinkCustomFields() {
     ]);
   }
 
+  /**
+   * Checks if a payment plan id is a manual payment
+   * 
+   * @param string $paymentProcessorID
+   * @return bool
```
MiyaNoctem: Please add a blank line to separate @param's from @return

Eg.
```
/**
   * Checks if a payment plan id is a manual payment
   * 
   * @param string $paymentProcessorID
   *
   * @return bool
   */
```

CRM/MembershipExtras/Upgrader.php
```diff
@@ -220,6 +221,162 @@ private function createPeriodLinkCustomFields() {
...
+    foreach ($recurringContributions as $paymentPlan) {
+      if ($this->isManualPaymentPlan($paymentPlan['payment_processor_id'])) {
+        $lastInstalment = $this->getLastInstalmentForPaymentPlan($paymentPlan['id']);
+        $lineItems = $this->getLineItemsForContribution($lastInstalment['id']);
+
+        $this->copyLastInstalmentLineItemsToRecurContrib($paymentPlan, $lineItems);
+      }
+
+      $isMatch = (
+        $this->isManualPaymentPlan($paymentPlan['payment_processor_id']) &&
```
MiyaNoctem: Yoy won't need to do this check when you only bring in payment plans.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -220,6 +221,162 @@ private function createPeriodLinkCustomFields() {
...
+
+  /**
+   * Finds all payment plans and populate the offline recurring contribution
+   * line item for the payment plans using an offline payment processor
+   */
+  private function updatePaymentPlans() {
+    $recurContributions = $this->getAllPaymentPlans();
+
+    foreach ($recurringContributions as $paymentPlan) {
+      if ($this->isManualPaymentPlan($paymentPlan['payment_processor_id'])) {
```
MiyaNoctem: Instead of fetching all recurring contributions and determining if they're a manual payment plan or not later, it might be more efficient to just query for the ones that are a payment plan.

```
$manualPaymentProcessors = CRM_MembershipExtras_Service_ManualPaymentProcessors::getIDs();
$result = civicrm_api3('ContributionRecur', 'get', [
  'payment_processor_id' => ['IN' => $manualPaymentProcessors],
  'options' => ['limit' => 0],
]);    
```

CRM/MembershipExtras/Upgrader.php
```diff
@@ -220,6 +221,162 @@ private function createPeriodLinkCustomFields() {
...
+        $lastInstalment = $this->getLastInstalmentForPaymentPlan($paymentPlan['id']);
+        $lineItems = $this->getLineItemsForContribution($lastInstalment['id']);
+
+        $this->copyLastInstalmentLineItemsToRecurContrib($paymentPlan, $lineItems);
+      }
+
+      $isMatch = (
+        $this->isManualPaymentPlan($paymentPlan['payment_processor_id']) &&
+        !empty($this->getLastInstalmentForPaymentPlan($paymentPlan['id'])) &&
```
MiyaNoctem: You're querying for the same contribution twice. Might be better to use the result in the variable and use it here, as `!empty($lastInstalment)`. Every call to the API to the DB is expensive, and since this needs to run for all existing payment plans, there could be a lot of them, so we need to try to be as efficient as possible.

MiyaNoctem: BTW, I'm not sure if this is why you're checking the las contribution here, but according to the ticket's description, you need to check if the number of instalments is more than 0. This is not the same as checking if there is at least one contribution for the payment plan, as payment plans with instalments = 0 may and probably will have a lot of contributions associated to it. It's weird, I know. Basically, this will happen if you don't choose a payment plan when you create the membership, but still end-up using a manual payment processor (like direct debit).

PR #84, opened 2018-10-12
BASW-279: Disable Tabs With Jquery API

16kilobyte:
## Overview
Currently, when clicking on any of those buttons, a .disabled-click class is applied to the tab's <a> elements. The style for that class (grayish color, no pointer cursor) is coming from the membershipextra's own stylesheet.

But the tabs are jQuery UI tabs, and those already provide an api which includes also a disable() method.

When you invoke that method, a .ui-state-disabled class is applied to the tabs, and that class works fine also in shoreditch.

## Before
Tabs were disable by adding the `disabled-click` class to <a> elements of all tabs.

## After
Tabs are being disabled by invoking the `disable()` method


guanhuan:
@16kilobyte looks good. Please attach screenshots for next period too.

16kilobyte:
@guanhuan Done

PR #86, opened 2018-10-16
BASW-222: Fix Add Membership Line to Standard Multi-line Membership Order (Next Period)

16kilobyte:
## Before
Issues reported
1. When "Add Membership" is clicked, the Financial Type column still says "select a membership type" even when an item from the dropdown has been selected - [Screenshot here](http://nimb.ws/cJfPs4)

2. When "Add Membership" or "Add Other Amount" is clicked, other buttons are still clickable even though they shouldn't be. - [Screenshot here](http://nimb.ws/Unc6FI)

3. "Add Membership" and "Add Other Amount" buttons are not compatible with Shoreditch

## After
Points above have been fixed.

### DIFF
--- a/js/NextPeriodLineItemHandler.js
+++ b/js/NextPeriodLineItemHandler.js
@@ -11,17 +11,19 @@ CRM.$(function () {
   });
 
   CRM.$('#next_buttons #addOtherAmount').on('click', function(e) {
-    e.preventDefault();
     CRM.$('#addLineItemRow').show();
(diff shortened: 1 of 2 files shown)


guanhuan:
@16kilobyte When a new row is being worked on, the tabs and the buttons in your gif still look enabled. I remember we fixed at least the tabs in another ticket if I am correct.

16kilobyte:
@guanhuan, this branch was branched off before the fix was merged.

PR #90, opened 2018-10-29
BASW-288: Record an Activity When Change is Made Installment Management Screen

16kilobyte:
## Overview
1. When an user adds/ removes a line from the current period, a "Update Payment Plan Current Period" type activity should be generated with:
a. Subject - "Update Payment Plan Current Period"
b. Source Record ID - ID of the recurring contribution
c. Added by - Admin
d. With - The contact of the recurring contribution

2. When an user adds/ removes a line from the next period, a "Update Payment Plan Next Period" type activity should be generated with:
a. Subject - "Update Payment Plan Next Period"
b. Source Record ID - ID of the recurring contribution
c. Added by - Admin
d. With - The contact of the recurring contribution

## Before
Action did not exist

## After
Action has been implemented
basw-288-after

### DIFF
--- a/js/NextPeriodLineItemHandler.js
+++ b/js/NextPeriodLineItemHandler.js
@@ -155,7 +155,7 @@ function showMembershipAddLineItemConfirmation() {
 
   CRM.confirm({
     title: ts('Add ' + membershipType.name + '?'),
-    message: ts('Please note the changes should take effect immediately after "Apply".'),
+    message: ts('Please note the changes should take effect immediately after \`Apply\'.'),
     options: {
       no: ts('Cancel'),
       yes: ts('Apply')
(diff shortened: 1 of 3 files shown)


guanhuan:
@16kilobyte 
1. I can't seem to find where you install those activity types. Could you point me where that's done?
2. Please attach gifs for the results.

js/NextPeriodLineItemHandler.js
```diff
@@ -222,7 +222,7 @@ function createLineItem(params) {
         null,
         'success'
       );
-
+      createActivity('Update Payment Plan Next Period', 'Update Payment Plan Next Period');
```
MiyaNoctem: I see the second parameter for the createActivity method is typeID. Are you sure 'Pdate Payment Plan Next Period' can be used as activity type ID? Shouldn't it be 'update_payment_plan_next_period'? Same for all other calls to that method.

16kilobyte:
@guanhuan Updated

PR #97, opened 2018-11-01
BASW-122: Fix Problem When Completing Payment

MiyaNoctem:
## Overview
When editing a pending contribution for a payment plan with a single instalment, and that payment plan corresponded to a membership with an overridden status, the recurring contribution was being set as Completed. This should not happen on payment plans with only one installment.

## Before
When editing a contribution for a payment plan membership with an overridden status, the system took a different path to update payment info, including a method run when a contribution is updated and its status is changed, affecting its recurring contribution:
https://github.com/civicrm/civicrm-core/blob/5.6.1/CRM/Contribute/BAO/Contribution.php#L242
https://github.com/civicrm/civicrm-core/blob/5.6.1/CRM/Contribute/BAO/ContributionRecur.php#L850

This method checked the recurring contribution and changed its status to 'Completed' if all installments had been paid.

## After
Fixed problem by implementing a pre hook on the Recurring Contribution entity, that calculates the plan's status according to our rules.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
+++ b/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
@@ -1,0 +1,158 @@
+<?php
+
+/**
+ * Implements pre hook on ContributionRecur entity.
+ */
+class CRM_MembershipExtras_Hook_Pre_ContributionRecur {
+
+  /**
+   * Operation being performe.
+   *
+   * @var string
+   */
+  private $operation;
+
+  /**
+   * Current data for the recurring contribution, if it's being updated.
+   *
+   * @var array
+   */
+  private $recurringContribution;
+
+  /**
+   * List of parameters that are being used to create/update the recurring
+   * contribution.
+   *
+   * @var array
+   */
+  private $params;
+
+  /**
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /**
+   * Loads the data for the given recurring contribution ID and returns it.
+   *
+   * @param int $id
+   *
+   * @return array
+   */
+  private function getRecurringContribution($id) {
```
omarabuhussein: I think it is better to move this close to where it first called, right after the constructor

CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
```diff
@@ -0,0 +1,158 @@
...
+   * @param array $params
+   */
+  public function __construct($op, $id, &$params) {
+    $this->operation = $op;
+    $this->recurringContribution = $this->getRecurringContribution($id);
+    $this->params = &$params;
+  }
+
+  /**
+   * Preprocesses the parameters ebing used to create or update the recurring
```
omarabuhussein: typo : ebing => being

PR #106, opened 2018-11-12
BASW-122: Fix Recurring Contribution End Date/Status Calculation on Completion

MiyaNoctem:
## Overview
When the full number of instalment contributions are marked as completed, the recurring contribution need to be marked as completed too. The recurring contribution end date should also be set to the cycle day of the last interval.

## Before
On completion of payment plan, end date was being set as actual date, instead of receive date of last installment of the payment plan.

## After
The recurring contribution end date is now set to the receive date of the last installment.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
+++ b/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
@@ -1,0 +1,158 @@
+<?php
+
+/**
+ * Implements pre hook on ContributionRecur entity.
+ */
+class CRM_MembershipExtras_Hook_Pre_ContributionRecur {
+
+  /**
+   * Operation being performed.
+   *
+   * @var string
+   */
+  private $operation;
+
+  /**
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
```diff
@@ -0,0 +1,158 @@
...
+    $isManualPaymentProcessor = in_array($this->recurringContribution['payment_processor_id'], $manualPaymentProcessorsIDs);
+
+    if ($isManualPaymentProcessor) {
+      return TRUE;
+    }
+
+    return FALSE;
+  }
+
+}
```
16kilobyte: Blank line at EOF

MiyaNoctem: It's there, it just looks like its's not. When tehre is no line at EOF, github shows it like this:

image

PR #111, opened 2018-11-20
BASW-322: Create Membership Period Entity

MiyaNoctem:
## Overview
Users should be able to recall individual membership periods on which a membership type was active for a contact.

We need to create a membership period entity as a sub entity of membership. The database table should have the following columns:

- id - primary key - required
- membership_id - foreign key to membership entity - required
- period_start_date - date when the membership period starts
- period_end_date - date when the membership period ends
- payment_entity_table - civicrm_contribution or civicrm_contribution_recur
- payment_entity_id - id of the payment entity
- is_active - whether the period has taken effect or not
- is_historic - whether it is an historical period created before the membership extras is installed

## Before
Entity did not exist.

## After
Generated new entity boilerplate code with civix.


CRM/MembershipExtras/BAO/MembershipPeriod.php
```diff
@@ -0,0 +1,27 @@
...
+   *
+   * @return CRM_MembershipExtras_DAO_MembershipPeriod|NULL
+   */
+  public static function create($params) {
+    $className = 'CRM_MembershipExtras_DAO_MembershipPeriod';
+    $entityName = 'MembershipPeriod';
+    $hook = empty($params['id']) ? 'create' : 'edit';
+
+    CRM_Utils_Hook::pre($hook, $entityName, CRM_Utils_Array::value('id', $params), $params);
+    $instance = new $className();
```
omarabuhussein: why to do this when you can just do : 

```
new CRM_MembershipExtras_DAO_MembershipPeriod();
```

api/v3/MembershipPeriod.php
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * MembershipPeriod.create API specification (optional)
+ * This is used for documentation and validation.
+ *
+ * @param array $spec description of fields supported by this API call
+ * @return void
+ * @see http://wiki.civicrm.org/confluence/display/CRMDOC/API+Architecture+Standards
+ */
+function _civicrm_api3_membership_period_create_spec(&$spec) {
+  // $spec['some_parameter']['api.required'] = 1;
```
omarabuhussein: please define any required item for the create API call in the specs here

sql/create_membership_period.sql
```diff
@@ -0,0 +1,22 @@
+SET FOREIGN_KEY_CHECKS=0;
```
omarabuhussein: maybe better to create a new directory under sql and rename the file so :  sql/Upgrader/0002_create_membership_period.sql

to make a better seperation

PR #113, opened 2018-11-22
BASW-323: Create Period on Membership Sign-up

MiyaNoctem:
## Overview
As a staff member, I want to record the period of each membership so that I can view the complete history of exact dates of signup, renewal and end, and any gaps in the membership where it is cancelled or paused.

- Every time a new membership is signed up, a new membership period should also be created under that membership.
- Period start date and period end date should be the membership start date and end date.
- Payment entity should be recurring contribution if paid via payment plan. Otherwise it should be contribution.

## Before
On membership creation, no period was created.

## After
Implemented a post hook on MembershipPayment entity, so that once all the data for membership and contribution is available, the period is created.

### DIFF
--- a/CRM/MembershipExtras/Hook/Post/MembershipPayment.php
+++ b/CRM/MembershipExtras/Hook/Post/MembershipPayment.php
@@ -67,7 +67,7 @@ class CRM_MembershipExtras_Hook_Post_MembershipPayment {
   public function postProcess() {
     if ($this->operation == 'create') {
       $this->fixRecurringLineItemMembershipReferences();
-      $this->updateMembershipPeriod();
+      $this->createNewMembershipPeriod();
     }
   }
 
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -38,6 +67,7 @@ public function __construct($operation, $objectId, CRM_Member_DAO_MembershipPaym
   public function postProcess() {
     if ($this->operation == 'create') {
       $this->fixRecurringLineItemMembershipReferences();
+      $this->updateMembershipPeriod();
```
omarabuhussein: createNewMembershipPeriod would be a more meaningful name I guess

CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -118,4 +141,45 @@ private function getRelatedRecurringLineItem() {
...
+  }
+
+  /**
+   * Checks if the period already exists for the payment entity.
+   *
+   * @return bool
+   */
+  private function periodExistsForMembershipPayment() {
+    $paymentEntityID = $this->contribution['contribution_recur_id'] ?: $this->contribution['id'];
+
```
omarabuhussein: no need for the space and I think it would be better to also check if the entity_type is correct

CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -118,4 +141,45 @@ private function getRelatedRecurringLineItem() {
...
+    ]);
+  }
+
+  /**
+   * Checks if the period already exists for the payment entity.
+   *
+   * @return bool
+   */
+  private function periodExistsForMembershipPayment() {
+    $paymentEntityID = $this->contribution['contribution_recur_id'] ?: $this->contribution['id'];
```
omarabuhussein: this line is used twice in this PR, you can move it to a new private method

PR #114, opened 2018-11-30
BASW-338: Allow users to view membership periods in membership tab

omarabuhussein:
## User Story

As a staff member, I want to see the period of each membership signup and renewal so that I can know the complete membership history of this contact. 
 Membership periods should be nested under each membership type where active periods are in black color, past periods are in gray color and inactive periods are in red, the periods should also be ordered by start date in Ascending order.

## Before

There is no place to view the current membership periods.

## After

Membership periods can be viewed by expanding the membership type field : 

+ On Legacy screen
cccaaaa

+ On Shoreditch screen
ddadasads

## Technical Notes 

- hook_civicrm_pageRun is used to add js and css files that created the nested view on membership tab.
-  The JS file listen to datatables (since most civicrm tables are created using [datatables plugin](https://datatables.net)) [init event](https://datatables.net/reference/event/init) all loops through all visible tables (which are active and inactive membership tables) and add click listener that shows/hides the membership periods nested view.
- The nested periods view is appended to the table using datatables child() method, more info can be found here  : https://datatables.net/examples/api/row_details.html

### DIFF
--- a/css/membershipPeriodsNestedView.css
+++ b/css/membershipPeriodsNestedView.css
@@ -1,22 +1,42 @@
 .membership-period-collapse-icon:before {
-  content: '► ';
+  content: '\f054';
+  font-family: 'FontAwesome';
+  font-weight: normal;
+  margin-right: 13px;
+  color: #4d4d69;
   cursor: pointer;
 }
+
+
 
 .membership-period-expand-icon:before {
-  content: '▼ ';
+  content: '\f078';
+  font-family: 'FontAwesome';
+  font-weight: normal;
+  margin-right: 13px;
+  color: #4d4d69;
   cursor: pointer;
 }
 
-.periods-nested-view-no-right-border td {
-  border-right: 0 !important;
+.periods-nested-view-table td {
+  border-right: 1px !important;
   text-align: left;
 }
 
+.periods-nested-view-table th{
+  background:#f3f6f7 !important;
+}
+
 .membership-period-inactive td{
-  color: #FF0000 !important;
+  color: #cf3458 !important;
 }
 
 .membership-period-in-past td{
   color: #c2c2c2 !important;
+}
+
+.periods-table-container {
+  padding-top: 0 !important;
+  border: 1px !important;
+  border-color: #0a0a0a !important;
 }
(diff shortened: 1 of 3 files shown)


guanhuan:
I have updated the wireframe slightly after our phase 4 catchup call. Please have a look in the ticket. Nothing major.

Overall looks great! In case you want to apply the updates in a different PR, this is good to go.

jamienovick:
+1 looks good. 

1. Just a small comment that the screen appears to show the memberships for a moment on their own without the periods before the JS kicks in. Not sure whether on slower browsers this could be a bit confusing. Is there anything we can do to perhaps have a loading spinner and hide the section till everything is loaded?

2. I've mentioned this before and I don't think we've executed on the design pattern but for gift aid we have a much better visual solution for subtables - I'm not sure if we have a screenshot of this somewhere.
@jonscreat will know I've mentioned this many times? Can you find one?
Can we get it executed like that easily or is that problematic?

JonScreat:
Regarding Jamie's second point here is the expandable table design:
screen shot 2018-12-03 at 10 35 16
It can be seen on Invision here: https://projects.invisionapp.com/d/main#/console/10432973/302007692/preview

omarabuhussein:
@guanhuan 

I couldn't find any wireframe change to the UI in this PR

omarabuhussein:
>>> Just a small comment that the screen appears to show the memberships for a moment on their own without the periods before the JS kicks in. Not sure whether on slower browsers this could be a bit confusing. Is there anything we can do to perhaps have a loading spinner and hide the section till everything is loaded?

The thing is the that the memberships view itself is controlled by CiviCRM, but will see if there is anything possible to do.

jamienovick:
Whatever you can do would be great. Not a huge deal. You can see an example of the CSS in the gift aid extension of that helps @guanhuan may be able to send you a screenshot

omarabuhussein:
Okay so I created a new PR for this https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/115

that handles the nested rows in away that is more close to CiviCRM core, it also fix the delay in JS problem and I also adjusted the style to look mush closer to the photo Jon attached. It also now show a spinner when expanding any membership periods (in case fetching the periods takes time).

PR #115, opened 2018-12-04
BASW-338: Allow users to view membership periods in membership tab

omarabuhussein:
## Note

This PR is to replace https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/114 by following an approach that is more closer to how civicrm core handles nested rows.

## User Story

As a staff member, I want to see the period of each membership signup and renewal so that I can know the complete membership history of this contact. 
 Membership periods should be nested under each membership type where active periods are in black color, past periods are in gray color and inactive periods are in red, the periods should also be ordered by start date in Ascending order.

## Before

There is no place to view the current membership periods.

## After

Membership periods can be viewed by expanding the membership type field : 

+ On Legacy screen
with-shorr

+ On Shoreditch screen
with-shorrrr

## Technical Notes 

- A new getOrderedMembershipPeriods() method is added to MembershipPeriod BAO which accepts a membership ID and return the a DAO object that contains all the membership periods ordered by start date.

- A new page is added (CRM_MembershipExtras_Page_MembershipPeriod_PeriodsTable ) that show the list of  membership periods in a table given the membership id as a GET parameter.

- a new template at **templates/CRM/Member/Page/Tab.extra.tpl** is added, the template basicly extends the core CRM/Members/Page/Tab.tpl which is responsible for rendering the membership tab page, but hence that this new template does not override the core one since I use the .extra.tpl which basicly appened the content of the template to the core one as explained here : https://docs.civicrm.org/dev/en/latest/framework/templates/#appending-jquery-or-other-code-to-a-template

- The new extra template add JQuery code that adds the nested periods view to the membership tab, the lines here https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/72a4b27b07f1327343882317870af907462a07b0/templates/CRM/Member/Page/Tab.extra.tpl#L28-L47
is copied from js/crm.expandRow.js which is the civicrm core way to handle such nested views, it basically load the (CRM_MembershipExtras_Page_MembershipPeriod_PeriodsTable ) page and appends it to the memberships table as a nested row instead.

- membershipextras_civicrm_pageRun hooks is used to add a CSS file that styles the membership periods nested view.

### DIFF
--- a/CRM/MembershipExtras/BAO/MembershipPeriod.php
+++ b/CRM/MembershipExtras/BAO/MembershipPeriod.php
@@ -1,5 +1,4 @@
 <?php
-use CRM_MembershipExtras_ExtensionUtil as E;
 
 class CRM_MembershipExtras_BAO_MembershipPeriod extends CRM_MembershipExtras_DAO_MembershipPeriod {
 
@@ -20,6 +19,17 @@ class CRM_MembershipExtras_BAO_MembershipPeriod extends CRM_MembershipExtras_DAO
     CRM_Utils_Hook::post($hook, 'MembershipPeriod', $instance->id, $instance);
 
     return $instance;
+  }
+
+  public static function getOrderedMembershipPeriods($membershipId) {
+    $membershipPeriod = new self();
+    $membershipPeriod->membership_id = $membershipId;
+    $membershipPeriod->orderBy('start_date ASC');
+    if ($membershipPeriod->find() > 0) {
+      return $membershipPeriod;
+    }
+
+    return NULL;
   }
 
 }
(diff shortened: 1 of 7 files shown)


omarabuhussein:
@guanhuan @jamienovick can u take a look at the gifs and let me know if all good ?

omarabuhussein:
> Close enough
that's the best I could get 😄 

> Also looks like the delay in showing the triangle has been sorted.

Yup it is.

omarabuhussein:
@guanhuan I will merge this now, and we will discuss the wireframes adjustments when you come back since I couldn't find the ones related to this ticket.

PR #118, opened 2018-12-18
BASW-324: Create Membership Period on Renewals

MiyaNoctem:
## Overview
Every time a membership is renewed, a new membership period should also be created under that membership. Period start date should be today or a day after the last active period end date (whichever is later). Period end date should be period start date plus the length of a full term.

## Before
Periods were being created on Membership entity post hook. This had the problem that the dates for the membership were not updated by the time it was run, thus the new period had the same dates as the previous period.

## After
Period creation logic was moved to the MembershipPeriod BAO. This in turn is called after each different form of membership renewal is completed:

- After a membership is renewed using the "Renew" action in the contact's membership Tab.
- After a membership is renewed in the auto-renewal job.
- After a membership is renewed using a Contribution page.

## Notes
This PR has also two commits cherry-picked from phase 3 workstream fith a fix for the auto renewal job. See PR #116.


CRM/MembershipExtras/Hook/PostProcess/ContributionConfirm.php
```diff
@@ -0,0 +1,36 @@
...
+   */
+  public function __construct(CRM_Contribute_Form_Contribution_Confirm $form) {
+    $this->form = $form;
+  }
+
+  /**
+   * Postprocesses form to creaate membership periods.
+   */
+  public function postProcess() {
+    $membershipID = $this->form->_params['membershipID'];
```
omarabuhussein: would be better to use the CRM_Utils_Array::value in case the membershipID index does not exist

CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
```diff
@@ -230,6 +231,23 @@ public function run() {
...
+  private function createMembershipPeriods() {
+    $recurringLineItems = $this->getNewPaymentPlanActiveLineItems();
+
+    foreach ($recurringLineItems as $lineItem) {
+      $priceFieldValue = !empty($lineItem['price_field_value_id']) ? $this->getPriceFieldValue($lineItem['price_field_value_id']) : [];
+      if (!$this->isMembershipLineItem($lineItem, $priceFieldValue)) {
+        continue;
+      }
+
+      $existingMembershipID = $this->getExistingMembershipForLineItem($lineItem, $priceFieldValue);
```
omarabuhussein: I would rename this method to **getExistingMembershipIdForLineItem** to indicate that it return an id

CRM/MembershipExtras/BAO/MembershipPeriod.php
```diff
@@ -21,6 +21,127 @@ public static function create($params) {
...
+   * Returns an array with the information for the latest active period for the
+   * given membership.
+   *
+   * @param int $membershipID
+   *
+   * @return array
+   */
+  public static function getLastActivePeriod($membershipID) {
+    $membershipPeriod = new self();
+    $membershipPeriod->membership_id = $membershipID;
```
omarabuhussein: I might be mistaken but wouldn't the last active membership has the is_active Flag = TRUE  too ?

MiyaNoctem: Yes it would, but that's the idea, to obtain the last period that has is_active = TRUE, even if its the current one.

PR #122, opened 2019-01-03
BASW-330: Allow users to delete membership periods

omarabuhussein:


CRM/MembershipExtras/BAO/MembershipPeriod.php
```diff
@@ -252,4 +259,62 @@ private static function updateMembershipDates($membershipPeriod) {
...
+
+    return $membershipPaymentsCount;
+  }
+
+  public static function isLastPeriodOfMembership($periodId) {
+    $membershipPeriod = self::getMembershipPeriodById($periodId);
+
+    $periodsOfMembership = new self();
+    $periodsOfMembership->membership_id = $membershipPeriod->membership_id;
+    return $periodsOfMembership->find() === 1;
```
MiyaNoctem: I'm not really sure how this can be used to know if that is the last period of  membership. It'd only work if the membership hs one period. 

I think you should search for all active periods of the membership, sort by end_date DESC and see if the first one returned has the same ID as `$periodId`.

Eg.
```php
    $membershipPeriod = self::getMembershipPeriodById($periodId);

    $membershipPeriod = new self();
    $membershipPeriod->membership_id = $membershipId;
    $membershipPeriod->is_active = TRUE;
    $membershipPeriod->orderBy('end_date DESC');

    if ($membershipPeriod->find(TRUE) > 0) {
      return $membershipPeriod->id == $periodId;
    }
```

omarabuhussein: the "last period" means the only renaming period  (or "if the membership has only one period and the user is trying to remove it") but probably your comment above indicate the I need to rename the method to make it more clear so instead of isLastPeriodOfMembership I might change it to isTheOnlyPeriodOfMembership, does this sound good ?

MiyaNoctem: Yes, that'd be perfect!

PR #125, opened 2019-01-14
BASW-262: Create Recurring Line Items for Memberships

MiyaNoctem:
## Overview
 After running Auto-Renew job, membership is not getting renewed for a Payment Plan with Null installments (Created membership & Selected 'Contribution'), with the Payment Status as 'Completed' & Auto-Renew set as 'True'. There is no entry in 'membershipextras_subscription_line' Table also after creating membership.

## Before
Memberships paid for with a renewable contribution, but not a payment plan, were not having their recurring line items created, as on that process flow line items and contributions are created before the recurring contribution is.

## After
Fixed by moving recurring line items creation from LineItem entity Post hook to Contribution post hook,
which will create line items once recurring_contribution_id is set.


MiyaNoctem:
Close in favor of PR #124, as the current fix doesn't work for payment plans with more than one installment.

PR #127, opened 2019-01-17
BASW-222: Fix Duplicate Memberships on Next Period

MiyaNoctem:
## Overview
When clicking the auto-renew checkbox on a line item in current period that is unset, so that it gets added to next period, a notice with warning sign should appear with text "This membership type is already enrolled in next period." when there is another membershipextras_subscription_line in next period with the same membership type.

## Before
A membership type could be added to next period, and then, added to current period without auto-renew. If auto-renew is then set, it resulted line item to be duplicated in next period, as when the auto-renew check-box was set, the line item was added as a donation on next period, instead of a membership. So each time auto-renew check-box was se on, a new line item was added to next period.

## After
Fixed by implementing validation of membership type, so if a line item for that membership type already exists on next period, a waning message is shown and no action is taken. Also, if validation passes ok, current line item is just set to be auto-renewed and recurring contribution - membership relationship is restored.

Also fixed a problem where amount for new membership line items was being calculated by dividing full
price by number of installments, resulting in the NaN amount when dividing by null or 0. Fixed by setting number of installments as 1 if unset or 0.

### DIFF
--- a/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
+++ b/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
@@ -164,7 +164,9 @@ class CRM_MembershipExtras_Page_EditContributionRecurLineItems extends CRM_Core_
       'start_date' => ['IS NOT NULL' => 1],
     ];
 
-    if (!$this->contribRecur['installments'] || $this->contribRecur['installments'] <= 1) {
+    $installments = CRM_Utils_Array::value('installments', $this->contribRecur, 0);
+
+    if ($installments <= 1) {
       $conditions['end_date'] = ['IS NULL' => 1];
     }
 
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
```diff
@@ -164,7 +164,9 @@ private function getCurrentPeriodLineItems() {
       'start_date' => ['IS NOT NULL' => 1],
     ];
 
-    if (!$this->contribRecur['installments'] || $this->contribRecur['installments'] <= 1) {
+    $installments = CRM_Utils_Array::value('installments', $this->contribRecur, 0);
+
```
omarabuhussein: no need for the space

CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
```diff
@@ -182,7 +184,9 @@ private function getNextPeriodLineItems() {
       'is_removed' => 0,
     ];
 
-    if (!$this->contribRecur['installments'] || $this->contribRecur['installments'] <= 1) {
+    $installments = CRM_Utils_Array::value('installments', $this->contribRecur, 0);
+
```
omarabuhussein: same here, no need for the space

PR #128, opened 2019-01-21
BASW-256: Fix Line Item End Date Calculation on Payments Completion

MiyaNoctem:
## Overview
A problem was found when renewing a payment plan, as the end dates for the recurring line items where not set. Currently, these end dates are set by a post hook on the Financial Transaction entity. For some reason, under certain circumstances, this hook is not dispatched, as the flow to update info in CiviCRM core is different.

## Before
Editing a contribution and changing its status to completed takes a different path to recording a payment in CiviCRM. This caused that when the Financial Transaction Hook was called, the conditions for it to run were not being met, in order to calculate the recurring contribution's end date and status, as also the end date for the recurring line items.

## After
Fixed by moving setting of line item end date to ContributionRecur post hook.

### DIFF
--- a/CRM/MembershipExtras/Hook/Post/ContributionRecur.php
+++ b/CRM/MembershipExtras/Hook/Post/ContributionRecur.php
@@ -1,4 +1,5 @@
 <?php
+use CRM_MembershipExtras_Service_ManualPaymentProcessors as ManualPaymentProcessors;
 
 /**
  * Implements post-process hooks on ContributionRecur entity.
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/Post/ContributionRecur.php
```diff
@@ -0,0 +1,70 @@
...
+    $this->updateLineItemEndDates();
+  }
+
+  /**
+   * Updates end dates for line items associated to the recurring contribution.
+   */
+  private function updateLineItemEndDates() {
+    $contributionStatus = CRM_Contribute_PseudoConstant::contributionStatus($this->contributionRecurBAO->contribution_status_id, 'name');
+
+    if ($contributionStatus === 'Completed' && $this->contributionRecurBAO->installments > 1) {
```
omarabuhussein: is : 

" $this->contributionRecurBAO->installments > 1"

needed ? since even recur contiution where installments = NULL will have one installment which mean and one or more recur line items.

MiyaNoctem: Very needed! Line items for recurring contributions with 1 or less installments should NEVER have their end date set when payment is completed.

End dates for line items for payment plans with 1 or less installments are set on renewal... I know, this is BS, but, that's how it goes...

PR #137, opened 2019-02-08
BASW-222: Disable Auto-renew Checkbox and Re-use Lines Among Periods

MiyaNoctem:
## Overview
If a user tries to add a membership line item to the current period and that membership type is already on next period, the auto-renew checkbox should be disabled and forcibly set to true, before applying changes. Also, when the data is saved, instead of creating a new line item for the membership and disallowing setting it to auto-renew (as the membership type is already on next period), we should use the existing line item and update it so it appears both on current and next periods.

## Before
When adding a membership to current period with auto-renew, the system checked if the membership type was on next period and if so, issued a warning and disallowed the creation of the membership unless the auto-renew checkbox was unset.

## After
System checks if membership type is already on next period when selecting membership type, and if so, disables auto-renew checkbox and forcibly sets it to true.

When creating a membership line item for the current period of a payment plan we check if it exists for next period. If it does, we use that existing line, instead of creating a new one.

### DIFF
--- a/CRM/MembershipExtras/Form/RecurringContribution/AddMembershipLineItem.php
+++ b/CRM/MembershipExtras/Form/RecurringContribution/AddMembershipLineItem.php
@@ -185,6 +185,7 @@ class CRM_MembershipExtras_Form_RecurringContribution_AddMembershipLineItem exte
       'contribution_recur_id' => $this->recurringContribution['id'],
       'auto_renew' => TRUE,
       'is_removed' => 0,
+      'options' => ['limit' => 0],
       'api.LineItem.getsingle' => [
         'id' => '$value.line_item_id',
         'entity_table' => ['IS NOT NULL' => 1],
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Form/RecurringContribution/AddMembershipLineItem.php
```diff
@@ -140,6 +149,68 @@ protected function createRecurringLineItem() {
...
+   * @return array
+   */
+  private function getNextPeriodLineItems() {
+    $lineItems = array();
+
+    $params = [
+      'sequential' => 1,
+      'contribution_recur_id' => $this->recurringContribution['id'],
+      'auto_renew' => TRUE,
+      'is_removed' => 0,
```
omarabuhussein: add limit => 0 too to get all line items

js/CurrentPeriodLineItemHandler.js
```diff
@@ -693,6 +688,22 @@ CRM.RecurringContribution.CurrentPeriodLineItemHandler = (function($) {
       var taxRate = Math.round(taxAccount.tax_rate * 100) / 100;
       CRM.$('#newline_tax_rate', this.newMembershipRow).html(taxRate + ' %');
     }
+
+    var params = this.buildNextPeriodLineItemCallParameters();
```
omarabuhussein: also update buildNextPeriodLineItemCallParameters() to have limit => 0 to fetch all line items

PR #151, opened 2019-02-26
BASW-259: Fix Upgrader

16kilobyte:
## Overview
Upgrader fails and exits with the fatal error as reported by QA:
```
Feb 22 14:29:48  [info] $CRM_Queue_Page_AJAX_runNext_error = CiviCRM_API3_Exception: "Mandatory key(s) missing from params array: contact_id, frequency_interval"
#0 C:\MAMP\htdocs\Civi5.6\sites\default\files\civicrm\ext\uk.co.compucorp.membershipextras-master\CRM\MembershipExtras\Upgrader.php(330): civicrm_api3("ContributionRecur", "create", (Array:3))
#1 C:\MAMP\htdocs\Civi5.6\sites\default\files\civicrm\ext\uk.co.compucorp.membershipextras-master\CRM\MembershipExtras\Upgrader.php(353): CRM_MembershipExtras_Upgrader->createCustomValueForPaymentPlan("3")
#2 C:\MAMP\htdocs\Civi5.6\sites\default\files\civicrm\ext\uk.co.compucorp.membershipextras-master\CRM\MembershipExtras\Upgrader.php(402): CRM_MembershipExtras_Upgrader->updatePaymentPlans()
#3 C:\MAMP\htdocs\Civi5.6\sites\default\files\civicrm\ext\uk.co.compucorp.membershipextras-master\CRM\MembershipExtras\Upgrader\Base.php(72): CRM_MembershipExtras_Upgrader->upgrade_0001()
```

The problem is caused by [this method](https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/bc68179688e76229fd62046a62b642778cb6296e/CRM/MembershipExtras/Upgrader.php#L320) which wrongly tries to create a CustomValue for the `next_period` and `previous_period` custom fields.

## Before
Upgrader fails

## After
Bug fixed

### DIFF
--- a/CRM/MembershipExtras/Upgrader.php
+++ b/CRM/MembershipExtras/Upgrader.php
@@ -309,27 +309,14 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
     }
   }
 
-  private function getCustomFieldId($customGroupName, $customFieldName) {
-    return civicrm_api3('CustomField', 'getvalue', [
-      'return' => 'id',
-      'custom_group_id' => $customGroupName,
-      'name' => $customFieldName,
-    ]);
-  }
-
   private function createCustomValueForPaymentPlan($paymentPlanId) {
-    $nextPeriodCustomFieldId = $this->getCustomFieldId('related_payment_plan_periods', 'next_period');
-    $prevPeriodCustomFieldId = $this->getCustomFieldId('related_payment_plan_periods', 'previous_period');
+    $custom_fields = array (
+      'previous_period' => 0,
+      'next_period' => 0
+    );
+    $params = array('entityID' => $paymentPlanId, $custom_fields);
 
-    civicrm_api3('ContributionRecur', 'create', [
-      [
-        'id' => $paymentPlanId,
-        'custom_' . $nextPeriodCustomFieldId => 0,
-      ], [
-        'id' => $paymentPlanId,
-        'custom_' . $prevPeriodCustomFieldId => 0,
-      ],
-    ]);
+    CRM_Core_BAO_CustomValueTable::setValues($params);
   }
 
   /**
--- a/CRM/MembershipExtras/Upgrader.php
+++ b/CRM/MembershipExtras/Upgrader.php
@@ -176,51 +176,58 @@ class CRM_MembershipExtras_Upgrader extends CRM_MembershipExtras_Upgrader_Base {
    * Add Related Payment Plan Periods' Custom Fields
    */
   private function createPeriodLinkCustomFields() {
-    civicrm_api3('CustomGroup', 'create', [
+    $result = civicrm_api3('CustomGroup', 'get', [
+      'sequential' => 1,
       'name' => 'related_payment_plan_periods',
-      'title' => E::ts('Related Payment Plan Periods'),
-      'extends' => 'ContributionRecur',
-      'style' => 'Inline',
-      'collapse_display' => 1,
-      'weight' => 10,
-      'is_active' => 1,
-      'table_name' => 'civicrm_value_payment_plan_periods',
-      'is_multiple' => 0,
-      'collapse_adv_display' => 0,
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Upgrader.php
```diff
@@ -309,27 +309,14 @@ private function copyLastInstalmentLineItemsToRecurContrib($paymentPlan, $lineIt
...
-    civicrm_api3('ContributionRecur', 'create', [
-      [
-        'id' => $paymentPlanId,
-        'custom_' . $nextPeriodCustomFieldId => 0,
-      ], [
-        'id' => $paymentPlanId,
-        'custom_' . $prevPeriodCustomFieldId => 0,
-      ],
-    ]);
+    CRM_Core_BAO_CustomValueTable::setValues($params);
```
MiyaNoctem: Hmm, the way it was before kind of could have worked too, except on back-end, but the params array that was being sent is kind of weird. So this should work if you do this too:

```php
civicrm_api3('ContributionRecur', 'create', [
  'id' => $paymentPlanId,
  'custom_' . $nextPeriodCustomFieldId => 0,
  'custom_' . $prevPeriodCustomFieldId => 0,
]);
```

16kilobyte: This doesn't work as it should. It attempts to create a contribution with empty contact_id and frequency_interval which was causing the fatal error.

MiyaNoctem: Yeah, because there were 2 arrays being sent, which is something back-end can't deal with. But if yu just send the one array, with 'id' and 'custom_xxxx' fields, it should work ok.

Wrong:
```php
// This effectively would try to create a recurring contribution, as CiviCRM isn't able to find the 'id' field. There is also no contact_id, causing the error. This array is actually similar to sending an empty array into the API call, as CiviCRM back-end can't deal with that kind of input.
[
      [
        'id' => $paymentPlanId,
        'custom_' . $nextPeriodCustomFieldId => 0,
      ], [
        'id' => $paymentPlanId,
        'custom_' . $prevPeriodCustomFieldId => 0,
      ],
    ]
```

Right:
```php
// ID is given, sent as a key within the array, uses that ID to update the existing recurring contribution.
[
  'id' => $paymentPlanId,
  'custom_' . $nextPeriodCustomFieldId => 0,
  'custom_' . $prevPeriodCustomFieldId => 0,
]
```

CRM/MembershipExtras/Upgrader.php
```diff
@@ -176,51 +176,58 @@ private function removeCustomExternalIDs() {
...
-          'is_active' => 1,
-          'is_view' => 1,
-          'is_selector' => 0,
-          'custom_group_name' => 'related_payment_plan_periods',
-          'column_name' => 'next_period',
-        ]
-      ],
     ]);
+
+    if ($result['count'] === 0) {
```
omarabuhussein: or : 

```
if ($result['count'] > 0) {
  return;
}
```

to limit the nesting

CRM/MembershipExtras/Upgrader.php
```diff
@@ -380,27 +399,35 @@ private function isMoreThanOneMonthOld($date = null) {
...
-      'label' => 'Update Payment Plan Next Period',
-      'api.OptionValue.create' => [
+    $optionValues = [
+      ['name' => 'update_payment_plan_next_period', 'label' => 'Update Payment Plan Next Period'],
+      ['name' => 'update_payment_plan_current_period', 'label' => 'Update Payment Plan Current Period'],
+    ];
+    foreach ($optionValues as $optionValue) {
+      $result = civicrm_api3('OptionValue', 'get', [
+        'sequential' => 1,
+        'name' => $optionValue['name'],
```
omarabuhussein: add 'option_group_id' => 'activity_type', since there might be option values with same names across different option values.

16kilobyte: Right.

PR #162, opened 2019-03-13
BASW-212: Fix JS Errors Adding/Renewing Lines On Current Period

MiyaNoctem:
## Overview
When adding  new donation to a payment plan not set to auto renew, tax rate was not being shown.

## Before
Some variables that were defined on next period tab were needed for some events happening on current tab, like renewing and adding a new donation. 

## After
Fixed by adding the variables to current tab and using them within the current period handler class.

### DIFF
--- a/templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
+++ b/templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
@@ -1,5 +1,6 @@
 <script>
   var currentFinancialTypes = JSON.parse('{$financialTypes|@json_encode}');
+  var recurringContributionID = {$recurringContributionID};
 
   {literal}
   CRM.$(function () {
--- a/templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
+++ b/templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
@@ -1,12 +1,12 @@
 <script>
-  var currentFinancialTypes = JSON.parse('{$financialTypes|@json_encode}');
   var recurringContributionID = {$recurringContributionID};
+  var financialTypes = {$financialTypes|@json_encode};
 
   {literal}
   CRM.$(function () {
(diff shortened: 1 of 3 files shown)


templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -1,5 +1,4 @@
 <script>
-var recurringContributionID = {$recurringContributionID};
 var financialTypes = JSON.parse('{$financialTypes|@json_encode}');
```
16kilobyte: `currentFinancialTypes` from Current Period tab should still be available here. No need for this line.

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -1,5 +1,4 @@
...
 var membershipTypes = JSON.parse('{$membershipTypes|@json_encode}');
```
16kilobyte: Same thing applies here as well.
`var membershipTypes = {$membershipTypes|@json_encode};` should do

templates/CRM/MembershipExtras/Page/NextPeriodTab.tpl
```diff
@@ -1,5 +1,4 @@
...
 var recurringContribution = JSON.parse('{$recurringContribution|@json_encode}');
```
16kilobyte: Same here.
`var recurringContribution = {$recurringContribution|@json_encode};`

templates/CRM/MembershipExtras/Page/CurrentPeriodTab.tpl
```diff
@@ -1,5 +1,6 @@
 <script>
   var currentFinancialTypes = JSON.parse('{$financialTypes|@json_encode}');
```
16kilobyte: This statement `JSON.parse('{$financialTypes|@json_encode}');` is redundant
`var currentFinancialTypes = {$financialTypes|@json_encode}` alone is sufficient

PR #163, opened 2019-03-17
RSE-80: Membership Period Rules

16kilobyte:
## Overview
**User Story**: As a staff member, I want to be able to keep memberships and membership periods up to date so that I can maintain an accurate membership history.

Create a settings page that allows staff to select which of the following two rules should be automatically applied on membership periods:
1. a number of days after a payment is overdue, a membership period should be deactivated (default option)

2. a number of days after a payment is overdue, a membership period end date should be adjusted to one of the following days with ability to offset a number of days

- overdue payment receive date

- today

## Before
Action did not exist

## After
rse-80-after-1


CRM/MembershipExtras/Form/MembershipPeriod/PeriodRules.php
```diff
@@ -0,0 +1,132 @@
+<?php
+
+use CRM_MembershipExtras_ExtensionUtil as E;
+
+/**
+ * Membership Period Rules form controller
+ * 
+ * @see http://wiki.civicrm.org/confluence/display/CRMDOC43/QuickForm+Reference
+ */
+class CRM_MembershipExtras_Form_MembershipPeriod_PeriodRules extends CRM_Core_Form {
```
MiyaNoctem: I think we should move this to a separate path, out of MembershipPeriod, as the form deals with settings, and not the MembershipPEriod directly. Maybe `CRM_MembershipExtras_Form_Settings_PeriodRules`?

CRM/MembershipExtras/Job/OverdueMembershipPeriodProcessor.php
```diff
@@ -13,6 +13,12 @@ class CRM_MembershipExtras_Job_OverdueMembershipPeriodProcessor {
    * @throws \Exception
    */
   public function run() {
+    $autoUpdateOverdueMembershipEnabled = CRM_MembershipExtras_SettingsManager::shouldUpdateOverdueMembership();
+    if (!$autoUpdateOverdueMembershipEnabled) {
+      return TRUE;
+    }
+
+    $autoUpdateOverdueMembershipEnabled = CRM_MembershipExtras_SettingsManager::shouldUpdateOverdueMembership();
```
MiyaNoctem: This line can be deleted, right? I mean, it's exactly the same as ln 16, $autoUpdateOverdueMembershipEnabled should still have the same value.

templates/CRM/MembershipExtras/Form/MembershipPeriod/PeriodRules.tpl
```diff
@@ -0,0 +1,95 @@
+<div class="region region-content">
```
MiyaNoctem: I think I'm liking even more having a 'Settings' path for settings forms, that way we can put there both period rules settings form and membership extras settings form. Not to be done now, we could refactor it later, but it would be cool if you can move this form to `templates/CRM/MembershipExtras/Form/Settings/PeriodRules.tpl` for `CRM/MembershipExtras/Form/Settings/PeriodRules.php`.

PR #165, opened 2019-03-21
RSE-82: Merge uk.co.compucorp.civicrm.members-only-events With uk.co.compucorp.membershipextras

vinuvarshith:
## Overview
This PR merges uk.co.compucorp.civicrm.members-only-events using
BASW-46-registering-another-participant-members-only-event branch onto uk.co.compucorp.membershipextras dvelop branch.

## Details
I have moved settings, config, install files to appropriate locations on this ext. Then copied and renamed helper functions in extension's .php file, checked and copied over logic within hooks (created hooks if not already there)
I have tested this in my local on a basic level.


membershipextras.php
```diff
@@ -333,6 +337,25 @@ function membershipextras_civicrm_pageRun($page) {
       'page-header'
     );
   }
+
+  $f = '_' . __FUNCTION__ . '_' . get_class($page);
```
MiyaNoctem: Interesting, but on this extension we're trying to follow a more Object Oriented approach, and delegate hook implementations to separate classes, to try to declutter this file and have each class deal with a single responsibility. A similar approach is possible with OO too, though. I think we should refactor all those functions that implement hooks in members-only extension into individual classes, but that might be something we need to double-check on with PM's, as I know this might require quite some time.

membershipextras.php
```diff
@@ -386,3 +409,419 @@ function membershipextras_civicrm_alterContent(&$content, $context, $tplName, &$
...
+    //Insert this tab into position 4 (after `Online Registration` tab)
+    $tabs = array_merge(
+      array_slice($tabs, 0, 4),
+      $tab,
+      array_slice($tabs, 4)
+    );
+  }
+}
+
+/* Helper functions for membersonlyevent */
```
MiyaNoctem: Yeah, like these kind of "helper" functions we should try to move into appropriate classes.

templates/CRM/Event/Page/members-event-button.tpl
```diff
@@ -0,0 +1,30 @@
...
+ | at info[AT]civicrm[DOT]org. If you have questions about the        |
+ | GNU Affero General Public License or the licensing of CiviCRM,     |
+ | see the CiviCRM license FAQ at http://civicrm.org/licensing        |
+ +--------------------------------------------------------------------+
+*}
+<div class="action-link section members_event_link-section register_link-{$snippet.position}">
+  <a href="{$snippet.url}" title="{$snippet.button_text}" class="button crm-members_event-button">
+    <span>{$snippet.button_text}</span>
+  </a>
+</div>
```
MiyaNoctem: Missing \n at EOF

omarabuhussein:
This is big change and I don't think it is a change in the right direction anyway, I will close this PR and remove this branch

PR #170, opened 2019-04-03
BASW-414: Moving period creation logic from form level hooks to pre and post entity hooks and correcting the calculation logic 

omarabuhussein:
## Overview

While this PR handle several issues with our previous period creation logic, the main motivation behind it is that it was only work on form level hooks (mainly postProcess hook on certain forms), but we still need periods to be created for membership created by webform, API .. etc. 

## Implementation details

1- All the postProcess hook period creation logic is removed in this commit  : https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/d33566644ea4a3eb8364020fa114ecbdb9a334c5
since we no longer creating periods at form level.

2- Now for the new period creation logic itself, it can be found in the following parts : 

A- **Membership post create hook** : This is self explained, when a membership is created, we create a period with 
=> start date =  membership join date
=> end date = membership end date (it will be empty for **lifetime** memberships)

Any period created by this hook will be active by default and it will not be linked to any payment entity since no payment information is available at this stage of membership creation (though we attach it later in MembershipPayment post create hook).

The code that handle this can be found in this commit : 
https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/3a8329bcdb3c6b5b2d4f6b3549f82ec82cff758f

B- **Membership pre edit hook** : 

This hook is triggered if we either updating the dates of membership manually, or if we renewing a membership since renewing a membership at the end result in altering the membership dates which triggers this hook or if we completing a pending membership contribution since completing a pending contribution will extend the membership. 

Now for editing the membership dates manually, we later are not going to allow users to do that (will have another ticket for that) and it will only be available  through the membership importer or through membership bulk update , in these cases there are the following scenarios that are currently handled by this hook : 

  1- **new membership join date >= first activated period end date** : an error will be triggered in this 
  case since we don't know what to do with the first activated period in this case and it is not correct to 
  just remove it automatically specially if it is linked to a payment entity.
  2- "new membership end date <= last active period start date : similar to the case above but for the 
  end date, since we will not know what to do with the last activated period.
  3- **new membership end date <  last active period date AND > last active period start date** : Here 
  we adjust the last active period end date to match the new membership end date
  4- **new membership join date > first active period start date AND < first active period end date** : 
  Here we adjust the first active period start date date to match the new membership join date.
  5- **new membership end date > last active period end date** : a new period will be created with start 
  date = last active period end date plus 1 day and with end date = the membership new end date. the 
  period will be active by default and it will not be linked to any payment entity since it is generated by 
  and edit process where no payment is involved.
  6- **new membership join date < first active period start date** : similar to case 5 but for the join date, 
  a new membership period will be created with start date = the new join date and with end date = first 
  active period start date minus 1 day.

Now in case of renewal, this hook will only be triggered if we are renewing with "completed" contribution or payment plan, and the reason is that renewing with pending payment will not extend the membership end date which means that there is no edit operation will happen on the membership.

Now if we renewing a membership with "completed" payment then in most cases it will just extend the end date, which will trigger the same scenario as in case 5 above where a new period will be created the represent the new membership term.

Finally if we are completing a renewal  pending contribution , then the membership should already have an inactive period (how this period is created will be explained later in point C below) and completing this contribution will actually result in adjusting (extending) the membership end date to equal the inactive period end date, in this hook we only make sure that the inactive period become active, which is something we can figure out by the presence of the $contributionId parameter.
There is an edge case though which happen if we are completing a pending contribution for an expired membership, in this case civicrm will adjust the start date of the membership to today date and the end date to 1 term from today date, so in this case we adjust the existing  period dates to match the new membership dates.

C- **MembershipPayment post  create hook** : 

If a MembershipPayment create hook is triggered then it means that there is a membership created or renewed using either "contribution" or "payment plan" option (it can also be achieved too by creating a webform with certain configurations).

Now if we actually created a new membership, then the membership post create hook should already catch that and create a period, but since before we did not have any payment information the period will not have any payment entity linked to it, but in this stage we should already have the relevant payment entity details so we just attach it to the already created period, we know which period should we attach the payment to by storing the created period in static variable : https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/d992f47d7564f1a70e5793658a99ef484bc2e842/membershipextras.php#L236-L239
so if the period static variable is set at the time of MembershipPayment creation, then we know that this is a membership creation event and the period created by this event is the one created inside membership post create hook.

Now in case of renewal we also do the same as described above by attaching the recurring contribution details to the period, but the main thing we use this hook for is if we renewing with a "pending" contribution or payment plan, since civicrm does not extend the membership in 'pending' payment case this means that Membership post edit hook will not be triggered and not period will be created, and this is when this hook comes into the play, this hook check the contribution status related to this membership payment and if it is pending it will know for sure that there is no period is already created, so it will create new inactive period that match the new membership term. this period can be activated if we complete the contribution attached to it as explained in Membership post edit hook section above.

D- **Contribution post edit hook** : 

IF we are paying a membership with 'payment plan' then the recurring contribution will be created by a logic we created in this extension, this means that the recurring contribution will be attached to the contribution separately (after the call to MembershipPayment hook), and because of that the created period will be linked to the contribution instead of the recurring contribution, in this hook we detect that by the presence of contribution_recur_id and adjust the period to use the recurring contribution instead.

3- We used to extend the membership end date if it renewed using a pending payment plan, but civicrm by default does not have a payment plan option (it is something we added in this extension) and there is only 'contribution' option and if we renewed a membership using a pending contribution then its end date will not change until we complete the contribution (pay the contribution). To ensure consiency in behavior with payment plan, I removed all the code that extend pending payment plan memberships upon renewal and the membership now will only be extended either upon the completing of the contribution, or completing the first contribution in payment plan case. 

This is done in these 3 commits : 

https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/f087b3a092aaa12f58fccd972497054d181d9cd3
https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/4f36ff636fdad8b752e28743910d467b8305d828
https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/db499295fe797de4fc179356de23dd90bac5a0b6

4- We also no longer care much about the start date of the membership and most of our period creation care about the join date since it is more accurate representation to the membership dates and while we most likely will hide it later in another ticket form UI, for now I added some validation on membership bulk update in this commit :https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/ab39da500b0e292ca7381d3cd6bfd8039518bcd9 
to ensure that while bulk editing memberships, both start and join date should be available and they should equal each others.

5- Also since lifetime memberships does not have an end date,  I ensured in this commit https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/170/commits/b6822f498aea6587beb853e92e1f7e0f892038c3
that only one period is created for a lifetime membership and whenever the join date of the membership is adjust it will adjust this period start date.


CRM/MembershipExtras/Hook/Pre/MembershipPeriodUpdater.php
```diff
@@ -0,0 +1,322 @@
...
+    }
+    elseif (!empty($this->membership['end_date'])){
+      $membershipEndDate = date('Ymd', strtotime($this->membership['end_date']));
+    }
+
+    $this->calculatedMembershipEndDate = $membershipEndDate;
+  }
+
+  public function process() {
+    $this->rectifyPendingPeriodUponPaymentCompletion();
```
MiyaNoctem: Maybe `activatePendingPeriodUponPaymentCompletion()` is clearer?

membershipextras.php
```diff
@@ -227,10 +227,26 @@ function membershipextras_civicrm_post($op, $objectName, $objectId, &$objectRef)
     $lineItemPostHook->postProcess();
   }
 
+  if ($objectName == 'Membership' && $op == 'create') {
+    $membershipPaymentPostHook = new CRM_MembershipExtras_Hook_Post_MembershipCreate($objectRef);
+    $membershipPaymentPostHook->process();
+  }
+
```
MiyaNoctem: Extra \n

CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -153,4 +176,150 @@ private function getRelatedRecurringLineItem() {
...
+      return;
+    }
+
+    $contributionStatus = $this->contribution['contribution_status'];
+    if ($contributionStatus != 'Pending') {
+      return;
+    }
+
+    $isPaymentPlanPayment  = !empty($this->recurringContribution) && !empty($this->recurringContribution['installments']);
+    if($isPaymentPlanPayment) {
```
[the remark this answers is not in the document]
omarabuhussein: the issue with this is that it will always call isFirstPaymentPlanContribution() and do the API
 call query no matter if it is a payment plan or not, I would prefer to avoid that, we have already many things that affect the performance and I would like to minimize that if it is possible.

CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -153,4 +176,150 @@ private function getRelatedRecurringLineItem() {
...
+    }
+
+    $currentStartDate->add(new DateInterval($interval));
+    $currentStartDate->sub(new DateInterval('P1D'));
+    return $currentStartDate->format('Ymd');
+  }
+
+  /**
+   * Since periods are sometimes created before
+   * the arability of any payment record, we here
```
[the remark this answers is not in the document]
omarabuhussein: I am inventing a new English vocabulary apparently 😅

CRM/MembershipExtras/Hook/Pre/MembershipPeriodUpdater.php
omarabuhussein: it does not only activate the period but also might update its date in some cases as mentioned in rectifyPendingPeriodUponPaymentCompletion() method description, that's why i think rectify is a better term here.

PR #180, opened 2019-07-24
RSE-66: Adding/Implementing the (payments that should always activate membership) setting

omarabuhussein:
A new setting is added called "**Payment methods that always activate memberships**"

2019-07-24 19_57_41-Payment Plan Settings _ ppphase3

which allow users to select one or more payment method.

If any membership is created or renewed with such payment method then it will activate the membership as well as the newly created membership period, and completing the pending payment after that will not result in extending the membership again.

The PR is divided into two parts, the first part add the setting itself and the code that manage its storage and UI: https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/180/commits/72cf1e142ed5db63037120e7831b2e72d7d23d88

and the other part implement the business logic around it : https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/180/commits/15a572ebdfb15039f7edbd684f0dab2463f70f78

and the 2nd Part, two classes are mainly affected : 

1- CRM_MembershipExtras_Hook_Post_MembershipPayment : creating a membership with payment will trigger this class, by checking the payment contribution payment_instrument_id field and checking if period attached to this payment is inactive, we can tell if the membership is created in pending status and if it created using a payment method that should activate the membership, and if both conditions apply we activate the inactive period by calling `MembershipPeriod::updatePeriod(['id' => $periodId, 'is_active' => true]);` which will automatically activate the period and the related membership as well as correct its dates and status.

2- CRM_MembershipExtras_Hook_Pre_MembershipEdit : by checking the previous and the current contribution statues we can tell if the membership edit is triggered by completing a pending or partially paid payment, and by checking the payment method and  if it match one of the configured payment methods then prevent CiviCRM from extending its end date.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
@@ -1,6 +1,6 @@
 <?php
 
-use CRM_MembershipExtras_Service_MembershipEndDateCalculator as MembershipEndDateCalculator;
+use CRM_MembershipExtras_SettingsManager as SettingsManager;
 
 /**
  * Implements hook to be run before a membership is created/edited.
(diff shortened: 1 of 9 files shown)


CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
```diff
@@ -156,13 +163,57 @@ private function getPaymentRecurringContributionID() {
...
+      unset($this->params['end_date']);
+    }
+
+  }
+
+  /**
+   * The current details of the completed contribution
+   * which is after its status and the other payment details
+   * are changed by the payment made.
+   * @return array
```
MiyaNoctem: There should be a \n between description and `@return`.

PR #182, opened 2019-08-08
RSE-62: Unify the activation/deactivation dialogs UI

omarabuhussein:
## Before

The activation/deactivation dialog looks different when doing it through the period edit form versus using the activation/deactivation quick-action buttons.

*Quick action button dialog*
screenshot-5

*edit form dialog*
screenshot-6

## After

UI on both dialogs look the same

ddddd


jamienovick:
@omarabuhussein It looks like we are still building this without shoreditch enabled? I thought we agreed to do it with Shoreditch so we didn't have style it again later? thanks

omarabuhussein:
@jamienovick the issue is the we have much trouble with achieving the functionality itself with all CivicRM quirks already to worry about how it to look on shoreditch (and all the trouble that come with shoreditch) .

but anyway, I will enable it before doing stuff on the extension next time. 

Please keep that in mind (@16kilobyte , @MiyaNoctem )

PR #183, opened 2019-09-02
RSE-80: Add help text

16kilobyte:
## Overview
The specs says help texts should be added to two fields on the settings page

## Before
rse-80-before-2

## After
rse-80-after-3
rse-80-after-4


jamienovick:
@16kilobyte can you add the details of the help text to the PR

16kilobyte:
@jamienovick I have updated the screenshots to show the added help texts

cc: @MiyaNoctem

jamienovick:
@16kilobyte looks like a typo in the error message:

- When a period is deactivated, it **will be** treated as **if** it has not taken effect

jamienovick:
This can be a negative number**.**

(add a full stop at the end).

With the help text do we explain why we have these two different options? It tells you they exist but not WHY they exist?

16kilobyte:
@jamienovick The PR was merged already. A new PR #184 to fix these typo issues.

PR #189, opened 2019-09-17
RSE-66: Fix Message After Updating Pending Contributions

MiyaNoctem:
## Overview
Message shown when updating a pending contribution for a payment method set to auto-complete was erroneously informing the user that the end date of the membership was changed, by a whole period.

## Before
The message was being generated deep within CiviCRM Core when processing the Contribution Form. It erronously took the updating of the contribution as an intent to renew the memberhip, and thus calculated a new end date by extending it for a full period. Periods remained unchanged, as they should, but the message still showed the wrong date.

The new end date is calculated here:
https://github.com/MiyaNoctem/civicrm-core/blob/master///CRM/Contribute/BAO/Contribution.php#L1908-L1912

The message is built here:
https://github.com/MiyaNoctem/civicrm-core/blob/master/CRM/Contribute/Form/Contribution.php#L1736
https://github.com/MiyaNoctem/civicrm-core/blob/master//CRM/Contribute/BAO/Contribution.php#L5248-L5253

And it is added to status on session here:
https://github.com/MiyaNoctem/civicrm-core/blob/master/CRM/Contribute/Form/Contribution.php#L1777

## After
The building of the erroneous message and its addition to status in session happens all before postProcess hooks are called, so the solution uses a postProcess hoook on the Contribution form to process status messages and filter out the date.

### DIFF
--- a/CRM/MembershipExtras/Hook/PostProcess/ContributionForm.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/ContributionForm.php
@@ -1,0 +1,50 @@
+<?php
+
+/**
+ * Class CRM_MembershipExtras_Hook_PostProcess_ContributionForm.
+ *
+ * Post-processes contribution form used to create, update and delete
+ * contributions.
+ */
+class CRM_MembershipExtras_Hook_PostProcess_ContributionForm {
+
+  /**
+   * Form object that is being processed.
+   *
+   * @var \CRM_Contribute_Form_Contribution
+   */
+  private $form;
+
+  /**
+   * CRM_MembershipExtras_Hook_PostProcess_ContributionForm constructor.
+   *
+   * @param \CRM_Contribute_Form_Contribution $form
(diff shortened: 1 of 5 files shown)


CRM/MembershipExtras/Hook/PostProcess/ContributionForm.php
```diff
@@ -0,0 +1,50 @@
...
+   */
+  public function __construct(CRM_Contribute_Form_Contribution &$form) {
+    $this->form = $form;
+  }
+
+  /**
+   * Post-processes the form.
+   */
+  public function postProcess() {
+    $isEditAction = $this->form->getAction() & CRM_Core_Action::UPDATE;
```
omarabuhussein: I move change this to : 

```
public function postProcess() {
   $this->fixMembershipEndDateInRenewalNotificaiton();
}

and move everything in postProcess to this new method.
```

PR #192, opened 2019-09-25
BASW-289: Do not save end date for lifetime meberships.

16kilobyte:
## Overview
On our instalment management screen, we are able to specify start date and end date when adding a membership line. The start date and end date will control two things:
1. narrowing down the instalments that the new membership line should apply to
2. setting the start date and end date of the membership

We want to avoid setting the end date of life memberships but also not taking away the ability to perform point 1 and 2 for other memberships.

## Before
basw-289-before-1

## After
basw-289-after
basw-289-after-2


jamienovick:
@omarabuhussein cc @16kilobyte should we not avoid merging this feature to develop until after the ASE testing? I dont think it's essential and may have risks of regressions no? @mikon

omarabuhussein:
hence that we reverted this PR here : https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/195

and that's why I am keeping the branch for this PR

PR #201, opened 2019-10-04
CGP-1466: Fix Taxes Applied Twice to Renewal Contributions

MiyaNoctem:
## Overview
Renewal of payment plans resulted in the new contributions having wrong amounts, as if taxes had been applied twice.

## Before
Problem was caused by a pre hook on the Contribution entity to fix contribution amount from all line items associated to the contribution. We implemented this on this PR #157 to fix contribution's amount when adding new line items to the payment plan. 

However, adding amounts when they were not being updated caused line items to be re-calculated by CiviCRM core, using the total amount of the contribution (ie. with tax) as the total for the line item, and calculating tax again from it.

## After
Fixed by only fixing contribution totals if set.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/Contribution.php
+++ b/CRM/MembershipExtras/Hook/Pre/Contribution.php
@@ -54,19 +54,29 @@ class CRM_MembershipExtras_Hook_Pre_Contribution {
    * Checks if total amount is ok vs sum of line items.
    */
   private function rectifyAmountsBasedOnLineItems() {
-    $lineItems = $this->getContributionLineItems();
-    $totalAmount = 0;
-    $taxAmount = 0;
+    $givenTotalAmount = CRM_Utils_Array::value('total_amount', $this->params, NULL);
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Hook/Pre/Contribution.php
```diff
@@ -54,19 +54,29 @@ public function preProcess() {
...
-      $totalAmount += $line['line_total'] + $lineTax;
-      $taxAmount += $lineTax;
+      $calculatedTotalAmount += $line['line_total'] + $lineTax;
+      $calculatedTaxAmount += $lineTax;
     }
-
-    if ($totalAmount != $this->params['total_amount'] || $taxAmount != $this->params['tax_amount']) {
-      $this->params['total_amount'] = $totalAmount;
-      $this->params['tax_amount'] = $taxAmount;
+    if ($calculatedTotalAmount != $givenTotalAmount || $calculatedTaxAmount != $givenTaxAmount) {
```
omarabuhussein: I don't think this condition is relevant and can be removed , if these amout does indeed match then assigning them to the params won't affect anything,

PR #203, opened 2019-10-07
CGP-1450: Tag Stable Release

MiyaNoctem:
## Overview
We need to update info.xml to tag first stable release of membership extras extension.

### DIFF
--- a/info.xml
+++ b/info.xml
@@ -8,11 +8,11 @@
     <author>Kacper Warda</author>
     <email>kacper@compucorp.co.uk</email>
   </maintainer>
-  <releaseDate>2018-01-30</releaseDate>
+  <releaseDate>2019-10-07</releaseDate>
   <version>1.0.0</version>
-  <develStage>alpha</develStage>
+  <develStage>stable</develStage>
   <compatibility>
-    <ver>4.7</ver>
+    <ver>4.17.4</ver>
   </compatibility>
   <comments/>
   <civix>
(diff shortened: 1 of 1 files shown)


info.xml
```diff
@@ -8,11 +8,11 @@
...
     <email>kacper@compucorp.co.uk</email>
   </maintainer>
-  <releaseDate>2018-01-30</releaseDate>
+  <releaseDate>2019-10-07</releaseDate>
   <version>1.0.0</version>
-  <develStage>alpha</develStage>
+  <develStage>stable</develStage>
   <compatibility>
-    <ver>4.7</ver>
+    <ver>4.17.4</ver>
```
omarabuhussein: I think there were another version that we need to test against, have any tests done on it ? if so it should be included here.

info.xml
```diff
@@ -8,11 +8,11 @@
     <author>Kacper Warda</author>
     <email>kacper@compucorp.co.uk</email>
   </maintainer>
-  <releaseDate>2018-01-30</releaseDate>
+  <releaseDate>2019-10-07</releaseDate>
   <version>1.0.0</version>
-  <develStage>alpha</develStage>
+  <develStage>stable</develStage>
```
omarabuhussein: I think this is more of a "beta"

info.xml
ritatorok: The version should be 5.17.4.

PR #209, opened 2019-10-10
CGP-1450: Fix Renewal of Old Plans With Multiple Installments

MiyaNoctem:
## OVerview
After upgrading, payment plans with multiple installments were not upgrading, even if the apparently met criteria to renew.

## Before
When upgrading, a new custom group, payment_plan_periods, is added to the recurring contribution entity. For existing payment plans, records for this new custom group are created, and initialized by setting the 'next_period' and 'prev_period' fields to 0. This caused a problem when upgrading old plans, as only plans with next_perio = NULL where being picked for auto-renew.

Problems in upgrader can cause recurring line items to be duplicated, if two simultaneous upgrade jobs are run, for example. These line items then cause exceptions to be thrown when renewing the payment plan, due to a unique key in the line items table.

## After
Implemented fix so payment plans with next_period = 0 are also picked for renewal. Also made upgrader idempotent when creating line items for recurring contributions and checking if line item already exists before creating one on renewal.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
@@ -46,7 +46,7 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_MultipleInstallmentPlan extend
           ccr.contribution_status_id != {$cancelledStatusID} 
           AND ccr.contribution_status_id != {$refundedStatusID}
          )
-         AND ppp.next_period IS NULL
+         AND (ppp.next_period IS NULL OR ppp.next_period = 0)
          AND msl.auto_renew = 1
          AND msl.is_removed = 0
          AND msl.end_date IS NOT NULL
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
@@ -761,11 +761,11 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
    */
   private function isDuplicateLineItem($lineItem) {
     $result = civicrm_api3('LineItem', 'get', [
-      'entity_table' => $lineItem['entity_table'],
-      'entity_id' => $lineItem['entity_id'],
-      'contribution_id' => $lineItem['contribution_id'],
-      'price_field_value_id' => $lineItem['price_field_value_id'],
-      'price_field_id' => $lineItem['price_field_id'],
+      'entity_table' => CRM_Utils_Array::value('entity_table', $lineItem),
+      'entity_id' => CRM_Utils_Array::value('entity_id', $lineItem),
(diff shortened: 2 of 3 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -46,7 +46,7 @@ protected function getRecurringContributions() {
           ccr.contribution_status_id != {$cancelledStatusID} 
           AND ccr.contribution_status_id != {$refundedStatusID}
          )
-         AND ppp.next_period IS NULL
+         AND (ppp.next_period IS NULL OR ppp.next_period = 0)
```
omarabuhussein: should SingInstallmentPlan be updated too, Also I think it is better if we update the upgrader to set the value to 'Null' too at the same time.

CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
```diff
@@ -741,4 +745,34 @@ protected function recordPaymentPlanFirstContribution() {
...
+   *
+   * @return bool
+   *   TRUE if it finds a line item with the same combination of fields, FALSE
+   *   otherwise.
+   *
+   * @throws \CiviCRM_API3_Exception
+   */
+  private function isDuplicateLineItem($lineItem) {
+    $result = civicrm_api3('LineItem', 'get', [
+      'entity_table' => $lineItem['entity_table'],
```
omarabuhussein: what of any one of these values is empty ,  these keys need to be checked before used.

CRM/MembershipExtras/Upgrader.php
```diff
@@ -335,6 +335,15 @@ private function getLineItemsForContribution($contributionId) {
    * @param array $lineItems
    */
   private function copyLastInstalmentLineItemsToRecurContrib($paymentPlan, $lineItems) {
+    /*
+     * If there is an exception thrown at some point, we need to make this
+     * idempotent, so line items are not duplicated if the upgrader is ran more
+     * than once.
+     */
+    if ($this->linesExistForPaymentPlan($paymentPlan)) {
+      $this->deleteLineITemsForPaymentPlan($paymentPlan);
```
omarabuhussein: why to delete then recreate instead of just skipping the below `(return;)`

omarabuhussein:
Maybe in next version of this extension I think we will need to use work queue to manage this scheduled job, It should be more robust to avoid duplicate running scheduled jobs issues.

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
MiyaNoctem: Single installment plans only have one recurring contribution. Each time it gets renewed, a new contribution is added to the same recurring contribution. Thus for single installment payment plans, there is no previous or next periods (these are actually references to the prior recurring contribution and the next recurring contribution).

MiyaNoctem:
Ran tests locally, upgrade and renewal ran smoothly.

PR #212, opened 2019-10-17
CGP-1488: Remove Restriction to Upgrade Old Plans Only

MiyaNoctem:
## Overview
Payment plans renewing within the next month are skipped from updating, as the idea was they would get renewed by normal upgrader, while all other plans would not be touched. This has changed and now we need to set period links for all plans, making sure only the last recurring contributions of each membership are renewed.

## Before
Only plans with end date older than a month were processed.

## After
All payment plans are processed.

### DIFF
--- a/CRM/MembershipExtras/Upgrader/Setup/PaymentPlanUpdater.php
+++ b/CRM/MembershipExtras/Upgrader/Setup/PaymentPlanUpdater.php
@@ -34,12 +34,9 @@ class CRM_MembershipExtras_Upgrader_Setup_PaymentPlanUpdater {
       $lineItems = $this->getLineItemsForContribution($lastInstalment['id']);
       $this->copyLastInstalmentLineItemsToRecurContrib($paymentPlan, $lineItems);
 
-      $isManualPlanOlderThanAMonth = (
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Upgrader/Setup/PaymentPlanUpdater.php
```diff
@@ -34,12 +34,9 @@ public function run() {
...
       $this->copyLastInstalmentLineItemsToRecurContrib($paymentPlan, $lineItems);
 
-      $isManualPlanOlderThanAMonth = (
-        ($paymentPlan['installments'] > 0) && $paymentPlan['auto_renew'] &&
-        $this->isMoreThanOneMonthOld($paymentPlan['end_date'])
-      );
-      if ($isManualPlanOlderThanAMonth) {
-        $this->createCustomValueForPaymentPlan($paymentPlan['id'], $lineItems);
+      $isAutoRenewPlanWithMultipleInstallments = ($paymentPlan['installments'] > 0) && $paymentPlan['auto_renew'];
+      if ($isAutoRenewPlanWithMultipleInstallments) {
```
omarabuhussein: isMoreThanOneMonthOld() method is no longer needed, can u remove it

PR #213, opened 2019-10-18
CGP-1490: Prevent Cancelled Plans From Having Their Status Changed

MiyaNoctem:
## Overview
When running the upgrader (or the scheduled job - need to find out which one causing it) all the cancelled recurring contributions are changed to status Pending or In progress. 

## Before
When running upgrader, contributions are updated to record links between periods. Doing this was triggering a Pre hook which recalculated the payment plan's status by counting payed contributions, thus breaking cancelled payment plans and setting the as either pending or in progress.

## After
Fixed by only calculating recurring contribution status if current status is eithe Pending, In Progress or Completed.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
+++ b/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
@@ -50,13 +50,7 @@ class CRM_MembershipExtras_Hook_Pre_ContributionRecur {
     $this->operation = $op;
     $this->recurringContribution = $this->getRecurringContribution($id);
     $this->params = &$params;
-
-    $contributionStatuses = civicrm_api3('OptionValue', 'get', [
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
```diff
@@ -40,6 +50,13 @@ public function __construct($op, $id, &$params) {
     $this->operation = $op;
     $this->recurringContribution = $this->getRecurringContribution($id);
     $this->params = &$params;
+
+    $contributionStatuses = civicrm_api3('OptionValue', 'get', [
+      'option_group_id' => 'contribution_status',
```
omarabuhussein: I would add limit => 0 and move this to separate method.

PR #220, opened 2019-11-25
BASW-269: Fix Minor Styling Issues With ShoreDitch

MiyaNoctem:
## Overview
Some styling issues with shoreditch needed to be fixed:

1. When clicked on Add other amount on Current Tab, new line caused previous end-dates to wrap.
image

2. Current Period - 'Select' item & 'Start Date' fields seem to be vertically misaligned.
image

3. Add a gap after period start date on next period tab.
image

## Before
Styling problems were found on form.

## After
Styling issues fixed.

1. When clicked on Add other amount on Current Tab, new line caused previous end-dates to wrap.
image

2. Current Period - 'Select' item & 'Start Date' fields seem to be vertically misaligned.
image

3. Add a gap after period start date on next period tab.
image

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -18,7 +18,7 @@ div.period-dates {
 
 #currentPeriodLineItems tr.columnheader th,
 #nextPeriodLineItems tr.columnheader th {
-  white-space: nowrap;
+  line-height: 23px;
 }
 
 tr.disabled-row th,
--- a/css/style.css
+++ b/css/style.css
@@ -1,3 +1,7 @@
+div.period-dates {
+  margin-bottom: 10px;
+}
+
 .crm-container a:hover .crm-i.fa-check,
 .crm-container a:hover .crm-i.fa-times,
 .crm-container a:hover .crm-i.fa-trash {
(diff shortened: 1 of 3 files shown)


jamienovick:
@MiyaNoctem on the "fixed" screenshots the header title looks like it's got a lot of unnecessary line height between the rows? (i.e. between "renew" and "automatically"  Can we fix that. 

One more point (not sure if its for this issue) but the notifications (top right) come up behind the modal. Should they be in front?

omarabuhussein:
> One more point (not sure if its for this issue) but the notifications (top right) come up behind the modal. Should they be in front?

I think this happens when the notification is already on screen and then you open the Modal, so I think it is fine

MiyaNoctem:
@jamienovick I've fixed the issue with the wrapping table headers in the last commit:

image

Fixing the one about the notifications might be more of a challenge for me, plus I think it'd be more of something to do on shoreditch, to make notifications appear on top of modals.

omarabuhussein:
@MiyaNoctem I approved the PR but probably u need @jamienovick  approval too

jamienovick:
Thanks @MiyaNoctem sorry to be picky. Can we have a bit more spacing between the column headings? If that's too fiddly then don't worry but at the moment it's hard to read the different column headings as they look like one word. Should the whole modal be wider to fit all the different elements?

MiyaNoctem:
@jamienovick , no problem at all. I tried making the modal wider, but I think that is handled by CiviCRM core, so I finally just decreased the line-height for the table headers, and let them wrap, Here's what it looks like now:

image

MiyaNoctem:
@jamienovick , so do you think I can merge this PR now?

jamienovick:
Yes looks good thanks.

PR #221, opened 2019-11-29
MAE-168: Fix Removal of Recurring Line Items for Past Contributions

MiyaNoctem:
## Overview
On some payment plans, after removing membership line item/contribution line item from 'Manage Installments', it was not updating the contribution amount & membership tab.

## Before
For payment plans created in the past (say for a membership from 2000-01-01 to 2000-12-31), with pending installments, if a line item was removed and the end date not adjusted, end date was being set as the current date. When removing line itmes from pending contributions, the system was only removing them for pending contributions AFTER the end date.

MAE-157-Issue1

## After
Implemented a validation on selected/default end date, such if there are no pending installments found after the end date, an exception is thrown to alert the user.

MAE-168-Issue1 - After


omarabuhussein:
@MiyaNoctem can you add before and after Gifs

PR #223, opened 2019-12-10
BASW-65: Recalculate Membership Status on Creation of Payment Plan

MiyaNoctem:
## Overview
When a membership was being created with both start and end dates in the past, and the first contribution was marked as completed, the membership was created in an 'Expired' status, even if it had other installments pending. This meant it was not picked up by scheduled job to recalculate status, as that job skips expired memberships.

## Before
Our Hook to alter membership status needs to check if the membership is related to a payment plan, to then check if 'in_arrears' and 'not_arrears' events have been met. In order to do this, the membership, and its related recurring contribution and installments all have to be already stored in the database. 

When creating a membership, CiviCRM calculates its status before storing it into the database, and then calls the hook to alter the status (also before storing it in DB). This meant by the time our hook executes, not all required information is in the database and thus, the original status of 'Expired' is set for the membership.

## After
Added a call to CiviCRM API to recalculate membership status when payments are created for memberships, thus ensuring all required information is available in DB to determine if arrears related events have been triggered.

## Note
The MembershipPayment post hook had wrong line endings. I separated fixing the line endings and implementing the fix on different commits, to facilitate the review.


CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -38,9 +38,26 @@ public function __construct($operation, $objectId, CRM_Member_DAO_MembershipPaym
...
     if ($this->operation == 'create') {
       $this->fixRecurringLineItemMembershipReferences();
+      $this->recalculateMembershipStatus();
     }
   }
 
+  /**
+   * Recaalculates the status of the related membership.
+   *
+   * Recalculates the statos of the related membership to check if payment plan }
```
omarabuhussein: some typo : statos  => status


and there is an extra }

MiyaNoctem: Fixed!

PR #224, opened 2019-12-11
MAE-155: Change How Period Start and End Dates are Calculated

MiyaNoctem:
## Overview
Period start and end date that was shown on manage installments pop-up was taken from the recurring contribution. However these values are used internally to manage auto-renewals, in such a way that periods that have not been renewed do not have an end date, until they completely paid for. Also, plans with single installment never have an end date, despite the related memberships do. All this can cause confusion to the user.

Thus, we need to change the logic that shows period start and end dates (for both current and next periods), so that these are calculated on the fly when viewing the Manage Installments page.

## Before
- Current period start and end dates came from the recurring contribution.
- Next period start date was calculated by checking the recieve date of the last installment, and adding one installment frequency period.

## After
- Current period start and end dates are calculated from earliest membership start date and largest membership end date.
- Next period start date is calculated from current period end date + 1 day.

### DIFF
--- a/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
+++ b/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
@@ -42,6 +42,20 @@ class CRM_MembershipExtras_Page_EditContributionRecurLineItems extends CRM_Core_
   private $nextLineItemMembershipTypes = [];
 
   /**
+   * List of line items for current period.
+   *
+   * @var array
+   */
+  private $currentPeriodLineItems = [];
+
+  /**
+   * List of line items for next period.
+   *
+   * @var array
+   */
+  private $nextPeriodLineItems = [];
+
+  /**
    * @inheritdoc
    */
   public function __construct($title = NULL, $mode = NULL) {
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
```diff
@@ -182,25 +182,23 @@ public function run() {
...
-    $this->assign('periodStartDate', CRM_Utils_Array::value('start_date', $this->contribRecur));
-    $this->assign('periodEndDate', CRM_Utils_Array::value('end_date', $this->contribRecur));
-
     $currentPeriodLineItems = $this->getCurrentPeriodLineItems();
     $this->setCurrentLineItemMembershipTypes($currentPeriodLineItems);
 
     $nextPeriodLineItems = $this->getNextPeriodLineItems();
     $this->setNextLineItemMembershipTypes($nextPeriodLineItems);
 
+    $this->assign('periodStartDate', $this->getEarliestMembershipStartDate($currentPeriodLineItems));
```
omarabuhussein: since $currentPeriodLineItems is now passed a lot, maybe you can move it to class property instead

PR #225, opened 2019-12-11
MAE-159: Fix Wrong Period Dates Shown After Plan Renewal

MiyaNoctem:
## Overview
Issues were found with the dates shown on the View/Manage Future Installments screen.

After Completing All Payments:
- Current Period End Date: Updated to recieve date of last installment, which is not correct.

## Before
Period end date being shown on screen was end date of recurring contribution, which starts empty and is finally set when all installments are completed with the recieve date of the last installment.

After renewal, first, membership start date was not being updated to the start date of the new period. Secondly, end date shown on line items was the membership's current end date, which of course, after renewal, changed.

The following example is for a payment plan starting on 2018-01-01, ending on 2018-12-31.

### Previous period After Renewal
image
- Period end-date is shown as 31st December 2019. This is wrong, as end date for this period should be 31st December 2018. This happens because period end date is calculated by checking the largest membership end date, which is in fact the renewed memberhips end date (ie. 31st December 2019).
- Line item end-date is shown as 31st December 2019. This is also wrong, should be 31st December 2018. Also calculated from largest membership end-date.

### New Period on Renewal
image
- Period start date start date is shown as 1st January 2018, which is wrong, as it should be 1st January 2019. This happens because period start date is calculated from earliest membership start date, and membership start dates were not being updated on renewal job.

## After
It was decided, we should only allow to View/Modify Future Inatallments on the last recurring contribution in a succession of renewals. Thus we no longer have to deal with wrong dates being shown on prior periods after a renewal, as the link was removed from these.

Also, fixed renewals so that membership start date is updated to start date of the new period.

### New Period on Renewal
image

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
@@ -529,7 +529,7 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
       $existingMembershipID = $this->getExistingMembershipForLineItem($lineItem, $priceFieldValue);
 
       if ($existingMembershipID) {
-        $this->extendExistingMembership($existingMembershipID, $lineItem);
+        $this->extendExistingMembership($existingMembershipID, $lineItem['start_date']);
       } else {
         $existingMembershipID = $this->createMembership($lineItem, $priceFieldValue);
       }
@@ -603,13 +603,13 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
    *
    * @param int $membershipID
    *   ID of the membership to be extended.
-   * @param array $lineItem
-   *   Data with the line item the membership belongs to.
+   * @param string $startDate
+   *   New start date for the membership.
    */
-  private function extendExistingMembership($membershipID, $lineItem) {
+  private function extendExistingMembership($membershipID, $startDate) {
     $membership = new CRM_Member_DAO_Membership();
     $membership->id = $membershipID;
-    $membership->start_date = $lineItem['start_date'];
+    $membership->start_date = $startDate;
     $membership->end_date = MembershipEndDateCalculator::calculate($membershipID);
     $membership->save();
   }
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
@@ -529,7 +529,7 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
       $existingMembershipID = $this->getExistingMembershipForLineItem($lineItem, $priceFieldValue);
 
       if ($existingMembershipID) {
-        $this->extendExistingMembership($existingMembershipID);
+        $this->extendExistingMembership($existingMembershipID, $lineItem);
       } else {
         $existingMembershipID = $this->createMembership($lineItem, $priceFieldValue);
       }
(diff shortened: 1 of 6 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
```diff
@@ -602,10 +602,14 @@ private function createMembership($lineItem, $priceFieldValue) {
    * Extend membership identified by given ID.
    *
    * @param int $membershipID
+   *   ID of the membership to be extended.
+   * @param array $lineItem
+   *   Data with the line item the membership belongs to.
    */
-  private function extendExistingMembership($membershipID) {
+  private function extendExistingMembership($membershipID, $lineItem) {
```
omarabuhussein: since you only need the start_date for now, you can just pass it alone and rename the argument to $newStartDate

CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
```diff
@@ -302,21 +334,16 @@ private function calculateCurrentPeriodEndDate() {
...
     $earliestDate = null;
 
     foreach ($lineItems as $line) {
-      if ($line['entity_table'] != 'civicrm_membership') {
-        continue;
-      }
-
-      $membership = $this->getMembership($line['entity_id']);
-      $membershipStartDate = new DateTime($membership['start_date']);
+      $startDate = new DateTime($line['start_date']);
```
omarabuhussein: shouldn't you only check the membership line items  ?

CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
```diff
@@ -278,17 +280,47 @@ private function setNextLineItemMembershipTypes() {
...
+    if (!isset($this->contribRecur['end_date'])) {
+      return $this->getLargestMembershipEndDate($this->currentPeriodLineItems);
+    }
+
+    $numberOfInstallments = 1;
+    if (!empty($this->contribRecur['installments'])) {
+      $numberOfInstallments = $this->contribRecur['installments'];
+    }
+
+    $intervalLength = CRM_Utils_Array::value('frequency_interval', $this->contribRecur, 0) * $numberOfInstallments;
```
omarabuhussein: but this feels different to me from what we are trying to achieve, start and end date calculations should be only based on the membership dates according to our discussion and  it is not necessarily will match the recur contribution dates and interval

CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
MiyaNoctem: Before when we only looked at  membership start dates it made sense, but you could, for example, acquire a membership for only the first year and not renew it, but keep paying a donation, or an event in subsequent years. What start date would we use then, if there is no membership? This would make it work on those cases.

CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
MiyaNoctem: The thing is, after renwal, the original membership end_date will be gone for ever. For example, we acquire a membership from 2018-01-01 to 2018-12-31. We complete all payments and renew it. Now the membership end date will be 2019-12-31. We cannot show that end date when we look at the previous period, as the end date of that period is and will always be 2018-12-31. 

Bear in mind the end date we're recording on both the recurring contribution and the recurring lines is the last recieve date of a payement plan. So I thought it better to use the original start date of the period to calculate the original end date. We could also look at the current end date and subtract as many terms as necessary, but I thought that could be even harder, as we'd have to find out what term the user is viewing. Say a user has had an annual membership renew for 10 years, and we open the manage installments form for the 7th term... we'd have to cycle through all the terms until that one (7th) to know we're looking at the seventh, and use the current end date to calculate the end date of the seventh term...

MiyaNoctem: I've removed this logic, as we no longer need it, since we are not allowing to View/Modify Future Installments for prior recurring contributions in a payment plan.

PR #226, opened 2019-12-17
CGP-1486: Fix Changing of Membership End Dates on Payment Record

MiyaNoctem:
## Overview
When recording a payment, membership end dates is getting updated, as if the membership was being renewed. 

## Before
We had logic in place on the Membership Entity pre hook to prevent this from happening, but this relied on the fact that the contribution would be updated BEFORE the membership, as we needed the contribution ID in order to determine if the membership was for a manual payment plan and thus prevent membership extension.

## After
Added a condition to the Membership pre hook by checking if the membership is being updated through the record payment form, so that if a payment is being recorded, membership end date extension is also prevented.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
@@ -39,7 +39,7 @@ class CRM_MembershipExtras_Hook_Pre_MembershipEdit {
    * Preprocesses parameters used for Membership operations.
    */
   public function preProcess() {
-    if ($this->paymentContributionID) {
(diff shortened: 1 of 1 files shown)


MiyaNoctem:
> I am not an advocate of using form level code inside BAO/DAO level hook but it seems we have to do it like this or otherwise we won't be able to alter the membership dates.

Completely agree. Onlly did it because we were already doing it here:
https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/226/files#diff-e22c19002b74ec88cbc974741a3f7ff8L143-L145

PR #229, opened 2019-12-24
MAE-175: Ensure that the auto-renew flag is set correctly after webform submission

omarabuhussein:
Please check https://github.com/compucorp/webform_civicrm_membership_extras/pull/12 for more details

### DIFF
--- a/CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
@@ -123,14 +123,13 @@ class CRM_MembershipExtras_Hook_PostProcess_RecurringContributionLineItemCreator
   }
 
   private function calculateAutorenewalFlag($lineItemParams) {
-    if ($lineItemParams['entity_table'] == 'civicrm_membership') {
-      $autoRenew = TRUE;
-      if (empty($lineItemParams['api.Membership.get']['values'][0]['contribution_recur_id'])) {
-        $autoRenew = FALSE;
-      }
+    if ($lineItemParams['entity_table'] != 'civicrm_membership') {
+      return FALSE;
     }
-    else {
-      $autoRenew = $this->autoRenewAddOnContributions;
+
+    $autoRenew = TRUE;
+    if (empty($lineItemParams['api.Membership.get']['values'][0]['contribution_recur_id'])) {
+      $autoRenew = FALSE;
     }
 
     return $autoRenew;
--- a/CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
@@ -10,6 +10,8 @@ class CRM_MembershipExtras_Hook_PostProcess_RecurringContributionLineItemCreator
 
   private $previousPeriodFieldID;
 
+  private $calculateAutorenewalFlag = FALSE;
+
   public function __construct($recurContributionID){
     $this->recurContributionID = $recurContributionID;
     $this->previousPeriodFieldID = $this->getCustomFieldID('related_payment_plan_periods', 'previous_period');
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
```diff
@@ -30,6 +32,16 @@ private function setRecurContribution() {
     $this->recurContribution =  $recurContribution['values'][0];
   }
 
+  /**
+   * When called, it will force the class
+   * to calculate the auto-renew flag if
+   * it should be set or not when
+   * creating the line items.
+   */
+  public function forceAutorenewalFlagCalculation() {
```
MiyaNoctem: I'm not sure if exposing this method to webforms is the best approach. Seems like it'd increment coupling between webforms and membership extras. Why would webforms need to know that it has to call this method before calling the create method?

Do you think it could be better to do it the other way around and add a different class that the webform calls to create the line items? Something like `CRM_MemberhsipExtras_API_RecurringContributionLineItemCreator`, and have that class call `forceAutorenewalFlagCalculation()` instead?

Eg.
```php
class CRM_MemberhsipExtras_WebformAPI_RecurringContributionLineItemCreator {
  public function create($contributionRecurId) {
    $lineItemCreator = new CRM_MembershipExtras_Hook_PostProcess_RecurringContributionLineItemCreator($contributionRecurId);
    $lineItemCreator->forceAutorenewalFlagCalculation();
    $lineItemCreator->create();
  }
}
```

Maybe we shouldn't be allowing webforms to use membership extras classes directly, but rather provide a consistent API entry point, like a controller class that then instantiates the appropriate membership extras class and asks it to do the work. That way if we decide to change anything in the structure of membership extras classes, we won't have to worry about propagating the change into webforms module also.

Eg.
```php
$membershipExtrasAPI = new CRM_MembershipExtras_API_Controller();
$membershipExtrasAPI->createRecurringContributionLineItems($recurContributionID);
```

CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
```diff
@@ -99,4 +122,18 @@ private function getCustomFieldID($fieldGroup, $fieldName) {
...
 
+  private function calculateAutorenewalFlag($lineItemParams) {
+    if ($lineItemParams['entity_table'] == 'civicrm_membership') {
+      $autoRenew = TRUE;
+      if (empty($lineItemParams['api.Membership.get']['values'][0]['contribution_recur_id'])) {
+        $autoRenew = FALSE;
+      }
+    }
+    else {
+      $autoRenew = $this->autoRenewAddOnContributions;
```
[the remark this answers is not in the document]
omarabuhussein: ops, this was the name of `autoRenewAddOnContributions` class property before I changed it, good catch

CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
omarabuhussein: hmm, interesting!

but the thing is, the webform support module depends on membershipextras extension and not the other way around so that's why I thought it is maybe a better idea to provide generic methods inside the above class (such as forceAutorenewalFlagCalculation)  that any part of membershipextras product can use.

But probably in this case it is a better idea to create this CRM_MemberhsipExtras_WebformAPI_RecurringContributionLineItemCreator  class since webforms is an integral part of membershipextras and probably the main way external users will interact with the the system so it make sense maybe to give it special support.

Though I feel that I disagree with the last part in which you suggest instead to implement the API like this : 

```
$membershipExtrasAPI = new CRM_MembershipExtras_API_Controller();
$membershipExtrasAPI->createRecurringContributionLineItems($recurContributionID);
```

because I feel that it is so generic and we might end up with many different unrelated APIs end points that does not belong together.

So to conclude, I think I will go with (CRM_MemberhsipExtras_WebformAPI_RecurringContributionLineItemCreator ) class option here since it provide the necessary abstraction and at the same time prevent putting everything in the same place in the future.

MiyaNoctem: Cool beans!

PR #231, opened 2020-01-02
MAE-171: Set Join Date Equal to Start Date

MiyaNoctem:
## Overview
When a new membership line item is added on Manage installments, membership status is always showing 'New' though the contribution payment status is pending.

Also Membership Since Date defaults to today's date always though we have membership start & end dates in the past.

## Before
Join date was always being set as today. No special treatment was being done to check if payment plan was pending, to set Membership as Pending.

## After
Join date is now set equal to start date.

When creating line item, set the membership's status to pending and set the skip status calculation to 1, so it is stored with that status.

## Notes
Added commit to fix line endings to Alter/CalculatedMembershipStatus hook class before implementing new logic.


CRM/MembershipExtras/Hook/Alter/CalculatedMembershipStatus.php
```diff
@@ -69,52 +77,33 @@ public function alterMembershipStatus(&$calculatedStatus, $arguments, $membershi
     $this->membership = $membership;
     $this->calculationArguments = $arguments;
 
+    // If membership is being created, it should start in 'Pending' status.
```
omarabuhussein: hmm, I only meant that membership added through "manage installments" should be defaulted to "pending" since it is a special case, but we don't want every membership created to start pending  (what if I create a membership with one completed payment for example ?) .

I think the best and most simple approach here is just to update the api call that create the membership in RM/MembershipExtras/Form/RecurringContribution/AddMembershipLineItem.php  class so it always pass "Pending" as the membership status.

PR #235, opened 2020-01-08
MAE-188: Fix Start Dates for Lines Added with Different Recieve Date

MiyaNoctem:
## Overview
On Manage Installments screen, membership line start date should be the actual membership start dates and not the received date of the contribution.

While signing up membership using Webforms, received date of DD membership is set based on Payment collection run dates set on DD configuration. But it does affect the membership dates on Manage Installments which is not correct. 

Same issue, if membership is created using 'New membership' or 'Add Membership' option with a different received date, it updates the received date on Manage Installments as the membership start date.

## Before
When adding a membership, if the recieve date for the first installment was different to the start date of the membership, the dates shown on the Manage Future Installments Screen showed contribution recieve date as start date, instead of the membership's actual start date. This is beacuse the dates shown on the screen are calculated by checking the earliest start date of the plan's line items, which are all set to the recieve date of the first contribution.

## After
Fixed by using membership start dates for lines that are memberships when calculating earliest start date on the manage future installments form. Also added logic on line item creation to store the earliest membership's start date on line items start dates when the payment plan is created.

### DIFF
--- a/CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
@@ -58,8 +58,6 @@ class CRM_MembershipExtras_Hook_PostProcess_RecurringContributionLineItemCreator
     foreach ($lastContributionLineItems as $lineItemParams) {
       $this->createRecurLineItem($lineItemParams, $earliestStartDate);
     }
-
-    $this->updateRecurringContribution($earliestStartDate);
   }
 
   private function getLastContributionLineItems() {
@@ -141,13 +139,6 @@ class CRM_MembershipExtras_Hook_PostProcess_RecurringContributionLineItemCreator
       'line_item_id' => $lineItemCopy['id'],
       'start_date' => $lineStartDate,
       'auto_renew' => $autoRenew,
-    ]);
-  }
-
-  private function updateRecurringContribution($recurringContributionStartDate) {
-    civicrm_api3('ContributionRecur', 'create', [
-      'id' => $this->recurContributionID,
-      'start_date' => $recurringContributionStartDate,
     ]);
   }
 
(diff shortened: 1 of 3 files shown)


jamienovick:
@MiyaNoctem I'm a bit confused about the logic for the above.

How is the screen actually calculating the period start date? Which fields does it refer to specifically? Thanks

MiyaNoctem:
> How is the screen actually calculating the period start date? Which fields does it refer to specifically? Thanks

@jamienovick it was looking at the start date of each line item and using the earliest one for both the period and each line. 

We have 3 start dates in DB: membership start date (only for lines that are memberships), recurring line item start date, and recurring contribution start date. The recurring contribution start date and each recurring line item start date are set to the recieve date of the first installment. I've changed the logic in that screen so it checks membership start date instead of line item start date for line items that are memberships. Other lines (donations, events) will still use the line start date.

CRM/MembershipExtras/Hook/PostProcess/RecurringContributionLineItemCreator.php
```diff
@@ -103,11 +139,18 @@ private function createRecurLineItem($lineItemParams) {
     CRM_MembershipExtras_BAO_ContributionRecurLineItem::create([
       'contribution_recur_id' => $this->recurContributionID,
       'line_item_id' => $lineItemCopy['id'],
-      'start_date' => $this->recurContribution['start_date'],
+      'start_date' => $lineStartDate,
       'auto_renew' => $autoRenew,
     ]);
   }
 
+  private function updateRecurringContribution($recurringContributionStartDate) {
```
omarabuhussein: I don't think that this is in an alignment with what we are doing to do, the point was to separate the recur contribution/contributions from the membership dates, the recur contribution start date at the end might differ from the membership date but you here are forcing it to be the same as the earliest membership start date.

omarabuhussein: hence that we need the recur contribution start date to equal the (earliest membership start date +- cycle day) only upon renewal or offline renewal, but it should not be the case on creation (the user/admin might supply the start date of the payment plan) but this here does not account for that.

MiyaNoctem: Sounds good.

PR #237, opened 2020-01-14
MAE-178: Add Validation to Membership Update Form to Prevent Changing its Membership Type

MiyaNoctem:
## Overview
Changing a membership's type after it has been created is only allowed by CiviCRM if it is not associated to a recurring contribution (ie. the membership is not auto-renewable). However, we need to prevent this from happenin on non-renewable payment plans too, as changing the membership type will cause several problems on existing installments.

## Before
CiviCRM allowed changing the membership type of a membership so long as it was not auto-renewable.

## After
Added validation to membership edit form so that if membership is changed and it belongs to a pyament plan, an error is shown with the text: 
>This membership is part of an ongoing payment plan and cannot be edited directly. To modify the active memberships in this payment plan please go to the manage future instalments screen which can be found as an action on the most recent recurring contribution record. There you can add or remove memberships.

### DIFF
--- a/CRM/MembershipExtras/Hook/ValidateForm/MembershipUpdate.php
+++ b/CRM/MembershipExtras/Hook/ValidateForm/MembershipUpdate.php
@@ -84,12 +84,9 @@ class CRM_MembershipExtras_Hook_ValidateForm_MembershipUpdate {
   }
 
   /**
-   * Obtains payment processor ID used for the last membership payment.
+   * Obtains last payment used for the membership.
    *
-   * Returns tha las payment processor ID used to pay for the membership with a
-   * recurring contribution.
-   *
-   * @return mixed
+   * @return array
    *
    * @throws \CiviCRM_API3_Exception
    */
@@ -107,7 +104,11 @@ class CRM_MembershipExtras_Hook_ValidateForm_MembershipUpdate {
       ],
     ]);
 
-    return $contribution['values'][0];
+    if ($contribution['count'] > 0) {
+      return $contribution['values'][0];
+    }
+
+    return [];
   }
 
   /**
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Hook/ValidateForm/MembershipUpdate.php
```diff
@@ -0,0 +1,156 @@
...
+        'limit' => 1,
+        'sort' => 'contribution_id DESC',
+      ],
+      'return' => [
+        'contribution_id',
+        'contribution_id.contribution_recur_id',
+      ],
+    ]);
+
+    return $contribution['values'][0];
```
omarabuhussein: this will throw a notice if no record found, probably you need to add an if statement and return NULL if no result found

CRM/MembershipExtras/Hook/ValidateForm/MembershipUpdate.php
```diff
@@ -0,0 +1,156 @@
...
+      return TRUE;
+    }
+
+    return FALSE;
+  }
+
+  /**
+   * Obtains payment processor ID used for the last membership payment.
+   *
+   * Returns tha las payment processor ID used to pay for the membership with a
```
omarabuhussein: typo "las" => "last"

also the description is not correct for the method.

CRM/MembershipExtras/Hook/ValidateForm/MembershipUpdate.php
```diff
@@ -0,0 +1,156 @@
...
+   */
+  private function getRecurringContribution($recurringContributionID) {
+    if (empty($recurringContributionID)) {
+      return [];
+    }
+
+    return civicrm_api3('ContributionRecur', 'getsingle', [
+      'sequential' => 1,
+      'id' => $recurringContributionID,
+      'options' => ['limit' => 0]
```
omarabuhussein: you don't need the 'limit' option here

PR #245, opened 2020-01-24
MAE-195: Fix Calculation of Renewed Period Start Date

MiyaNoctem:
## Overview
Period start date for a renewed payment plan with multiple installments should be calculated by using the previous period end date and adding 1 day.

## Before
Period start date was being determined by calculating the recieve date of the nth+1 installment. Eg. on a payment plan with 12 installments, the new period's start date would've been the recieve date of what would be the 13th installment.

## After
Period start date is now alculated from the previos period's start date (largest membership end date) and adding one day. If a payment plan is being renewed without any memberships, start date of new perios id still calculated from nth+1 installment recieve date.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
@@ -9,6 +9,13 @@ use CRM_MembershipExtras_Service_MoneyUtilities as MoneyUtilities;
 class CRM_MembershipExtras_Job_OfflineAutoRenewal_MultipleInstallmentPlan extends CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
 
   /**
+   * List of line items to be renewed.
+   *
+   * @var array
+   */
+  private $linesToBeRenewed = [];
+
+  /**
    * Returns a list of payment plans with multiple installments that have at
    * least one line item ready to be renewed (ie. has an end date, is not
    * removed and is set to auto renew), mmeting these conditions:
@@ -99,6 +106,7 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_MultipleInstallmentPlan extend
     $paymentProcessorID = !empty($currentRecurContribution['payment_processor_id']) ? $currentRecurContribution['payment_processor_id'] : NULL;
 
     $this->paymentPlanStartDate = $this->calculateNewPeriodStartDate();
+    $this->membershipsStartDate = $this->calculateRenewedMembershipsStartDate();
     $paymentInstrumentName = $this->getPaymentMethodNameFromItsId($currentRecurContribution['payment_instrument_id']);
 
     $newRecurringContribution = civicrm_api3('ContributionRecur', 'create', [
@@ -141,6 +149,20 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_MultipleInstallmentPlan extend
    * @throws \Exception
    */
   private function calculateNewPeriodStartDate() {
+    $installmentReceiveDateCalculator = new InstallmentReceiveDateCalculator($this->currentRecurringContribution);
+
+    return $installmentReceiveDateCalculator->calculate($this->currentRecurringContribution['installments'] + 1);
+  }
+
+  /**
+   * Calculates the start date renewed memberships should have.
+   *
+   * @return string
+   *   Date the renewed memberships should have as start date.
+   *
+   * @throws \Exception
+   */
+  private function calculateRenewedMembershipsStartDate() {
     $latestDate = NULL;
     $currentPeriodLines = $this->getRecurringContributionLineItemsToBeRenewed($this->currentRecurContributionID);
     foreach ($currentPeriodLines as $lineItem) {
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -131,6 +132,63 @@ private function createRecurringContribution() {
...
+   * Calculates the new period's start date from the largest membership end date
+   * of the previous period.
+   *
+   * @return string
+   *   The new period's start date.
+   * @throws \Exception
+   */
+  private function calculateNewPeriodStartDate() {
+    $latestDate = NULL;
+    $currentPeriodLines = $this->getRecurringContributionLineItemsToBeRenewed($this->currentRecurContributionID);
```
omarabuhussein: I could be mistaken , but I see that getRecurringContributionLineItemsToBeRenewed() is called twice with the same recur contributions Id passed to it, so maybe its result can be cached and reused ? though would worth confirming first that it is the case

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -131,6 +132,63 @@ private function createRecurringContribution() {
...
+    foreach ($currentPeriodLines as $lineItem) {
```
omarabuhussein: I know it is generally better to use the API, but I think here it probably better if we optimize it by fetching and sorting the results with SQL and joins instead.

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -91,13 +91,14 @@ public function renew() {
...
+   *
+   * @throws \Exception
    */
   private function createRecurringContribution() {
     $currentRecurContribution = $this->currentRecurringContribution;
     $paymentProcessorID = !empty($currentRecurContribution['payment_processor_id']) ? $currentRecurContribution['payment_processor_id'] : NULL;
 
-    $installmentReceiveDateCalculator = new InstallmentReceiveDateCalculator($currentRecurContribution);
-    $this->paymentPlanStartDate = $installmentReceiveDateCalculator->calculate($currentRecurContribution['installments'] + 1);
+    $this->paymentPlanStartDate = $this->calculateNewPeriodStartDate();
```
omarabuhussein: the thing here is that the payment plan start date should be derived from the membership start date as you already doing, but it could differ based on cycle day. but you here setting them both to be the same

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -141,20 +149,32 @@ private function createRecurringContribution() {
...
     $currentPeriodLines = $this->getRecurringContributionLineItemsToBeRenewed($this->currentRecurContributionID);
     foreach ($currentPeriodLines as $lineItem) {
       if ($lineItem['entity_table'] != 'civicrm_membership') {
         continue;
       }
 
-      $membership = $this->getMembership($lineItem['entity_id']);
-      $membershipDate = new DateTime($membership['end_date']);
-
+      $membershipEndDate = new DateTime($lineItem['memberhsip_end_date']);
```
omarabuhussein: I think u need skip memberships without end date (lifetime memberships)

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -163,8 +183,7 @@ private function calculateNewPeriodStartDate() {
       return $latestDate->format('Y-m-d');
     }
 
-    $installmentReceiveDateCalculator = new InstallmentReceiveDateCalculator($this->currentRecurringContribution);
-    return $installmentReceiveDateCalculator->calculate($this->currentRecurringContribution['installments'] + 1);
+    return $this->calculateNewPeriodStartDate();
```
omarabuhussein: I think  since u already calculated this before and assigned it to this->paymentPlanStartDate, then here you can just do :

```
return this->paymentPlanStartDate;
```

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
```diff
@@ -247,7 +268,7 @@ private function copyRecurringLineItems($currentContribution, $nextContribution)
         CRM_MembershipExtras_BAO_ContributionRecurLineItem::create([
           'contribution_recur_id' => $nextContribution['id'],
           'line_item_id' => $newLineItem['id'],
-          'start_date' => $nextContribution['start_date'],
+          'start_date' => $this->paymentPlanStartDate,
```
omarabuhussein: my memory is a bit vague about this, but should the recur line item start date contain the related membership start date , or should  it really be the recur contribution start date as u doing here?

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
MiyaNoctem: I did something similar. Didn't think it was a good idea to couple those two methods, so instead I'm returning NULL on default, and handled it by checking if it returned NULL assigning $this->paymentPlanStartDate to memberships start date.

CRM/MembershipExtras/Job/OfflineAutoRenewal/MultipleInstallmentPlan.php
MiyaNoctem: It should be the date the recurring contribution starts, which has to do with the cycle day. 

On manage installments screen, the period start date is the earliest membership start date. On that same screen, each line shows the membersip start date, if it's a membership, recurring line start_date if not (eg. donations, events, etc).

On contributions tab, the recurring contribution start date is the date of the first installment (as affected by cycle day). On memberships tab, start date is the membership start date.

PR #255, opened 2020-03-31
MAE-227: Update Recurring Contribution Start Date on Cycle Day Change

MiyaNoctem:
## Overview
When Cycle Day of a recurring contribution is changed, its start date is not updated.

## Before
Changing the cycle day did not update the recurring contribution's start date.

## After
Changing the cycle day will recalculate the start date, based on the new cycle day.

## Notes
The file CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php had wrong line endings. I separated the line-endings fix in a different commit, and added a second commit with just the fixes done to update the recurring contribution's start date, to facilitate review.

### DIFF
--- a/CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
@@ -79,10 +79,7 @@ class CRM_MembershipExtras_Hook_PostProcess_UpdateSubscription {
       $params['next_sched_contribution_date'] = $this->nextContributionDate;
 
       $firstInstallment = $this->getFirstInstallment();
-      $isInstallmentPending = $firstInstallment['contribution_status'] == 'Pending';
-      $isReceiveDateInFuture = !$this->isReceiveDateInThePast($firstInstallment);
-
-      if ($isInstallmentPending && $isReceiveDateInFuture) {
+      if ($firstInstallment['contribution_status'] == 'Pending') {
         $params['start_date'] = $firstInstallment['receive_date'];
       }
     }
--- a/CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
@@ -75,9 +75,28 @@ class CRM_MembershipExtras_Hook_PostProcess_UpdateSubscription {
 
     if ($this->isUpdatedCycleDate()) {
       $params['next_sched_contribution_date'] = $this->nextContributionDate;
+      $params['start_date'] = $this->calculateNewStartDate();
     }
 
     civicrm_api3('ContributionRecur', 'create', $params);
+  }
+
+  /**
+   * Calculates the new start date for the recurring contribution.
+   *
+   * Uses cycle day to calculate the start date for the contribution.
+   *
+   * @return string
+   */
+  private function calculateNewStartDate() {
+    $formValues = $this->form->exportValues();
+    $currentStartDate = new DateTime($this->recurringContribution['start_date']);
+
+    $newStartDate = new DateTime(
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -75,11 +75,30 @@ private function updateRecurringContribution() {
...
 
+  /**
+   * Calculates the new start date for the recurring contribution.
+   *
+   * Uses cycle day to calculate the start date for the contribution.
+   *
+   * @return string
+   */
+  private function calculateNewStartDate() {
+    $formValues = $this->form->exportValues();
```
omarabuhussein: I left a note on the ticket saying that the frequency unit should also be taken into the account when doing the start date calculation from the cycle day.

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -1,296 +1,331 @@
...
+    civicrm_api3('ContributionRecur', 'create', $params);
+  }
+
+  /**
+   * Obteins the start date of the first contribution.
+   *
+   * @return string
+   */
+  private function getReceiveDateOfFirstInstallment() {
+    $contributions = $this->getContributions();
```
omarabuhussein, line 90: you don't need to get all the contribution to filter out the first one, just alter the method to get the first one directly.

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -1,296 +1,331 @@
...
+
```
omarabuhussein, line 91: also according to the  ticket new specs, the recur contribution start date should not necessarily get updated when updating the cycle day,  but only when all the contributions are in the future and are pending,

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -1,296 +1,331 @@
...
+   *
+   * @throws \Exception
+   */
+  private function updateContribution($contribution, $instrumentID, $installmentNumber) {
+    if ($contribution['contribution_status'] != 'Pending') {
+      return;
+    }
+
+    $params = [];
+    if ($this->isUpdatedCycleDate() && !$this->isReceiveDateInThePast($contribution)) {
```
omarabuhussein, line 268: typo : the method name should be isUpdateCycle**Day** not isUpdateCycle**Date**

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
```diff
@@ -1,296 +1,331 @@
...
+   * @param array $contribution
+   *
+   * @return bool
+   * @throws \Exception
+   */
+  private function isReceiveDateInThePast($contribution) {
+    $now = new DateTime();
+    $receiveDate = new DateTime($contribution['receive_date']);
+
+    if ($receiveDate < $now) {
```
omarabuhussein, line 298: you will probably need to rest the time here when comparing or otherwise you might get issues when the date = today date.

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
MiyaNoctem: Dates are now calculated using the dates calculator, which uses cycle day.

CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
MiyaNoctem: `$this->getContributions()` is also used when updating installments. So I've brought into this method the call to API to obtain only the first contribution.

PR #256, opened 2020-04-01
MAE-242: Stop Showing Payment Plan Toggler on Submit Card Membership

MiyaNoctem:
## Overview
The memberships tab of a contact record has a 'Submit Credit Card Membership' button. Clicking it will open a form to use live payment methods to add a membership. 'Payment Plan' option is displayed. This is not supported for Credit card membership and only 'Contribution' option should be displayed here. 

## Before
The same form is used by CiviCRM to show both the offline and on-line membership creation forms. The only thing that distinguishes onw from the other is the use of a 'mode' variable that, in the case of Credit Card Memberships, will hold the value 'live'.

image

## After
Fix by checking the mode of the form, so that if it is 'live', the payment plan toggle functionality is not added at the buildForm stage.

image


jamienovick:
@MiyaNoctem is it possible to write a test for this functionality? If not can you explain why not and keep track for our discussion? thanks

MiyaNoctem:
@jamienovick , I've added the test. Will have Omar check now.

PR #258, opened 2020-04-09
MAE-263: Prevent completing the contribution from the contribution edit form

omarabuhussein:
## Before

Users could complete a single contribution mainly using two methods, either from the contribution edit form or using the "Record payment" option, And since civicrm core runs different code for each method that is becoming sometimes annoying to support both ways, so we decided to minimize the ways in which users can complete contributions so they can now only use the "Record Payment" option to complete contributions.

aaaadd

 

## After
If the the contribution status is Pending, then you cannot change it to Completed.

dadsdsdas

### DIFF
--- a/CRM/MembershipExtras/Hook/BuildForm/ContributionEdit.php
+++ b/CRM/MembershipExtras/Hook/BuildForm/ContributionEdit.php
@@ -1,0 +1,13 @@
+<?php
+
+class CRM_MembershipExtras_Hook_BuildForm_ContributionEdit {
+
+  public function buildForm() {
+    $this->addPreventCompletingContributionJSFile();
+  }
+
+  private function addPreventCompletingContributionJSFile() {
+    CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.membershipextras', 'js/preventCompletingContribution.js');
+  }
+
+}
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/BuildForm/ContributionEdit.php
```diff
@@ -0,0 +1,31 @@
...
+  }
+
+  /**
+   * Appends another template to the
+   * core contribution edit form template
+   * to allow for modifications to
+   * core template.
+   */
+  private function appendModificationTemplate() {
+    CRM_Core_Region::instance('page-body')->add([
```
MiyaNoctem, line 26: Since this will only have javascript code, do you think it may be better to use something like this instead?

```php
CRM_Core_Resources::singleton()->addScriptFile('uk.co.compucorp.membershipextras', 'js/preventCompletingContribution.js');
```

templates/CRM/Contribute/Form/ContributionModifications.tpl
```diff
@@ -0,0 +1,12 @@
+<script type="text/javascript">
+  {literal}
+  preventChangingStatusFromPendingToCompleted();
```
MiyaNoctem, line 3: Do we need to wrap this in document.ready() event?

ie.
```js
CRM.$(function($) {
  preventChangingStatusFromPendingToCompleted();
});
```

PR #260, opened 2020-04-16
MAE-271: Fix Removal of Multiple Donation Items from a Plan

MiyaNoctem:
## Overview
Removal would only work if there was a single line that was not a membership, on the payment plan.

## Before
When there were multiple lines, the system did not have a reliable way to identify the line that should be removed, thus it chose to skip it, and no lines would be removed from any of the contributions.

## After
Fixed by adding more details to identify the line, and removing the first line found if several meet the criteria. Criteria being used to search for the line item was:

- 'entity_table',
- 'contribution_id'
- 'entity_id'
- 'price_field_id'
- 'price_field_value_id'

I've added:
- 'financial_type_id'
- 'qty'
- 'unit_price'

This way we can better zero-in on the line item that needs to be deleted. This may still yield more than one line item, but as discussed with @jamienovick, we can assume if lines meet all this criteria, they can be treated as equivalent, and thus any of the lines can be removed.

### DIFF
--- a/CRM/MembershipExtras/Form/RecurringContribution/RemoveLineItems.php
+++ b/CRM/MembershipExtras/Form/RecurringContribution/RemoveLineItems.php
@@ -291,7 +291,7 @@ class CRM_MembershipExtras_Form_RecurringContribution_RemoveLineItems extends CR
         'entity_id' => $entityID,
         'price_field_id' => $this->recurringLineItemData['price_field_id'],
         'price_field_value_id' => $this->recurringLineItemData['price_field_value_id'],
-        'financial_type_id' => $this->recurringLineItemData['price_field_value_id'],
+        'financial_type_id' => $this->recurringLineItemData['financial_type_id'],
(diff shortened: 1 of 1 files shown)


jamienovick:
@MiyaNoctem I don't think I said
"since as long as the financial type and amounts match, it doesn't matter what line is removed"

I said, that as long as all the data fields are the same. Are there any other fields that maybe different on the lines other than those two fields?

Thanks

MiyaNoctem:
@jamienovick ,

I was oversimplifying for the sake of brevity, as this is the complete list of attributes that need to match:

- 'entity_table',
- 'contribution_id'
- 'entity_id'
- 'price_field_id'
- 'price_field_value_id'
- 'financial_type_id'
- 'qty'
- 'unit_price'

But most of those fields were already being used before to search for the line item, I just had to add financial_type, qty and unit_price, as we did discuss the label should not be used. I've updated the PR description to leave this clearer, though.

PR #263, opened 2020-04-22
MAE-262: Update Readme With Details About CiviCRM Patches

MiyaNoctem:
## Overview
We need to update readme files with the details on CiviCRM patches, which ones we're maintaining and why.

## Before
Readme implied any ol' CiviCRM after v4.7.x could be used, which is not true for the new version of the extension.

## After
Updated Readme with a section explaining wht versions of patched CiviCRM can bes used, provided links to download, and explained the nature of each patch.

The current state (after changes) of the readme file can be seen here:

https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/262/readme.md

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -24,19 +24,44 @@ This will allow staff to modify a members benefits during the current membership
 CiviCRM has support for many payment processors, including several Direct Debit payment processors. With these “online” payment processors, when the membership comes to renew, the logic is actually managed by the payment processor in order to renew the membership and take next years payment. CiviCRM doesn’t however have any functionality for memberships where the payment is “offline” i.e. some Direct Debit processes or where you invoice clients in advance of receiving the payment. With Membership extras CiviCRM now fully supports offline automated renewal including sending email notifications with invoices for payment. We also have created a new offline batch direct debit export module which allows for full management of high volume direct debits through export processes.
 
 ## How do I get Membership Extras?
-Membership Extras is designed to work with CiviCRM 4.7.x or 5.x plus. If you are on an earlier version of CiviCRM, you will need to upgrade your site first or contact info@compucorp.co.uk if you needs assistance to do so.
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -24,19 +24,44 @@ This will allow staff to modify a members benefits during the current membership
...
 
 ## How do I get Membership Extras?
-Membership Extras is designed to work with CiviCRM 4.7.x or 5.x plus. If you are on an earlier version of CiviCRM, you will need to upgrade your site first or contact info@compucorp.co.uk if you needs assistance to do so.
+Membership Extras is designed to work with a patched version CiviCRM 5.19.4 plus. If you are on an earlier version of CiviCRM, you will need to upgrade your site first or contact info@compucorp.co.uk if you need assistance to do so.
```
omarabuhussein, line 27: we are currently testing on v5.24.4 , and it is the version we are supporting for next release , so I don't think  we need to include the others since there is no guarantee that they will still work.

readme.md
```diff
@@ -24,19 +24,44 @@ This will allow staff to modify a members benefits during the current membership
...
+
+#### Patch #2: Prevent Taxes to be Recalculated and Added to Membership Fee Each Time a Validation Fails
+There is a known bug in CiviCRM, tracked [here](https://lab.civicrm.org/dev/core/issues/778), where when creating a membership and clicking save, if a validation fails (say for example if the user skipped a required field), taxes would be recalculated and added again to the membership's fee. So, for example, if a membership's price was $100 with a 10% tax, the full price would be calculated to $110 before saving the membership. After the first validation error, the price would be recalculated to $121, then to $133.10 on the second failed validation, and so on every time the validation failed.
+
+Our patch fixes this by making sure every time the validation fails, the membership's fee without tax is used to calculate taxes. This is a temporary fix, though, as CiviCRM core team has laid down new groundwork on which we can base a more permanent solution that can be part of CiviCRM core in the near future.
+
+The PR with the bug fix can be seen [here](https://github.com/civicrm/civicrm-core/pull/15895).
+
+### If I am a user
```
omarabuhussein, line 56: I think we need to include that `Update Membership Statuses` scheduled job should be enabled.

PR #274, opened 2020-07-23
MAE-331: Add Linter, Test Actions

erawat:
### Overview

This PR adds

- Github Actions
  - Linters action to run phpcs linters again the php codes that have been added or updated when PR is created.
  - Test action to run phpunit tests to run tests against the repository when PR is created. 
- Github Pull Request Template
- PHPCS Ruleset 

This PR also changes the phpunit bootstrap level from classloader to settings, so we could fabricate the CiviCRM settings on unit tests.

### DIFF
--- a/.github/tests.yml
+++ b/.github/tests.yml
@@ -1,0 +1,52 @@
+name: Tests
+
+on: pull_request
+
+jobs:
+  run-unit-tests:
+
+    runs-on: ubuntu-latest
+    container: compucorp/civicrm-buildkit:1.0.0
+
+    env:
+      CIVICRM_EXTENSIONS_DIR: site/web/sites/all/modules/civicrm/tools/extensions
+
+    services:
+      mysql:
+        image: mysql:5.7
+        env:
+          MYSQL_ROOT_PASSWORD: root
+        ports:
+          - 3306
+        options: --health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3
+
+    steps:
+
+      - name: Config mysql database as per CiviCRM requirement
+        run: echo "SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));" | mysql -u root --password=root --host=mysql
+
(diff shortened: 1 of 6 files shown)


.github/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+    steps:
+
+      - name: Config mysql database as per CiviCRM requirement
+        run: echo "SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));" | mysql -u root --password=root --host=mysql
+
+      - name: Config amp
+        run : amp config:set --mysql_dsn=mysql://root:root@mysql:3306
+
+      - name: Build Drupal site
+        run: civibuild create drupal-clean --civi-ver 5.24.6 --web-root $GITHUB_WORKSPACE/site
```
omarabuhussein, line 32: I think we had a discussion about this in another PR but I kinda forgot, but we use a patched civicrm version and not the core one

.github/tests.yml
```diff
@@ -0,0 +1,52 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          repo: compucorp/civicrm-core
+          version: 5.24.6
+          path: site/web/sites/all/modules/civicrm
+
+      - uses: actions/checkout@v2
+        with:
+          path: ${{ env.CIVICRM_EXTENSIONS_DIR }}/uk.co.compucorp.manualdirectdebit
```
omarabuhussein, line 44: what this exactly do ?

erawat: @omarabuhussein  good caught. updated.

.github/tests.yml
erawat: @omarabuhussein  yes, we have patched action to apply our patched.  see https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/274/files#diff-ba72dd894f9b61a2c721a74670e376ccR34

omarabuhussein: oh right

PR #276, opened 2020-07-24
MAE-331: Renewal start date

erawat:
### Overview
This PR adds additional setting field called "Update start on renewal" to Payment plan settings and update the logic for creating membership start date based on the value selected on the setting when membership is auto-renewed via schedule job. 

If the setting is checked: 

- the start date of the membership should be set to the end date +1 day

if the setting is unchecked:

- the start date of the membership remains the same

The PR also includes some other codes changes which had errors reported by PHPCS.  

### Before

1. Update start on renewal was not present. 
Screenshot from 2020-07-24 17-22-51
2. When membership is auto-renewed by schedule job, the membership start date always set to the end date +1 day.

### After
1. Update start on renewal is added to the Payment plan settings page. 
Screenshot from 2020-07-24 17-23-48

2. When membership is auto-renewed by schedule job, the membership start date will be set to the end date +1 day only when the "Update start on renewal" on the Payment plan settings page is checked.  

### Technical Details

On calculateRenewedMembershipsStartDate() function, we check if the setting "Update start on renewal" not checked, if not we get the existing membership start date by getting membership id from line item then return existing membership start date immediately, so we ensure that the membership start date will remain the same. 

Unit test was not included in this PR yet as we are waiting for this https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/265 to be reviewed, so we don't duplicate fabricators, and we would follow the same practice that we would be using for Membership Extras extension. 

### Comment 

There is phpunit test fails on Github action on the existing test which is not relevant to this pull request.

### DIFF
--- a/CRM/MembershipExtras/Hook/Post/ContributionRecur.php
+++ b/CRM/MembershipExtras/Hook/Post/ContributionRecur.php
@@ -41,11 +41,11 @@ class CRM_MembershipExtras_Hook_Post_ContributionRecur {
     if ($contributionStatus === 'Completed' && $this->contributionRecurBAO->installments > 1) {
       $subscriptionLines = $this->getSubscriptionLines();
 
-      foreach($subscriptionLines as $line) {
+      foreach ($subscriptionLines as $line) {
         if (!empty($line['start_date']) && empty($line['end_date'])) {
           civicrm_api3('ContributionRecurLineItem', 'create', [
             'id' => $line['id'],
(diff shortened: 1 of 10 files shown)


omarabuhussein (approved):

omarabuhussein:
hence that some issues were caused by this new update which I fixed in this PR : https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/282

and merged to this PR branch, and since all passed testing I am merging them now.

PR #291, opened 2020-09-08
MAE-383: Change text on Installment to instalment on the payment plan form

erawat:
## Overview
This PR is change wording from Installment to Instalment.

## Before
Text was Installment.

## After
Text is not Instalment.

## Technical Details

This PR will only affect the User Interface only and the class, variable, database schema should be changed when the work need to be done.

### DIFF
--- a/CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
+++ b/CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
@@ -6,11 +6,11 @@
  */
 class CRM_MembershipExtras_Hook_BuildForm_MembershipPaymentPlan {
 
-  CONST DEFAULT_INSTALLMENTS_NUMBER = 12;
+  const DEFAULT_INSTALLMENTS_NUMBER = 12;
 
-  CONST DEFAULT_INSTALLMENTS_FREQUENCY = 1;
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Hook/BuildForm/MembershipPaymentPlan.php
```diff
@@ -53,16 +53,16 @@ private function addPaymentPlanSection() {
...
     $this->form->setDefaults(['installments' => self::DEFAULT_INSTALLMENTS_NUMBER]);
 
     $this->form->add('text', 'installments_frequency', ts('Interval'), '', FALSE);
-    $this->form->addRule('installments_frequency', ts('Installments must be a number.'), 'numeric');
+    $this->form->addRule('installments_frequency', ts('Instalments must be a number.'), 'numeric');
     $this->form->setDefaults(['installments_frequency' => self::DEFAULT_INSTALLMENTS_FREQUENCY]);
 
-    $this->form->add('select', 'installments_frequency_unit',
-      ts('Installments Frequency Units'),
+    $this->form->add('select', 'instalments_frequency_unit',
```
MiyaNoctem, line 64: Are you sure you want to change the field's name from `installments_frequency_unit` to `instalments_frequency_unit`? Wouldn't that require use to change the js in the tpl and the post-processing of the form?

MiyaNoctem (approved):

PR #292, opened 2020-09-08
MAE-378:  Update Payment plan block

erawat:
## Overview
This PR is for updating Payment Plan Block to hide below fields when payment plan is selected.

- Payment plan start date (received date)
- Amount (Membership amount)
- Payment status
- Financial type

The PR also moved Payment Method to the top before Record Payment from a Different Contact? field. 

## Before
Peek 2020-09-08 16-36

## After
Peek 2020-09-08 16-32


MiyaNoctem (commented):
Saw a submodule ` bin/civicrm/coder` and a binary file `bin/phpcbf` were added here. Do we really need this? If so, could you comment on the PR description why?

erawat:
> Saw a submodule ` bin/civicrm/coder` and a binary file `bin/phpcbf` were added here. Do we really need this? If so, could you comment on the PR description why?

@MiyaNoctem  that was a mistake, both files are removed. Thanks for spotting this.

MiyaNoctem (approved):

PR #295, opened 2020-09-18
IO-72: Create Automated membership upgrade rule create/update forms

omarabuhussein:
## Overview

This creates forms for creating and updating automated membership upgrade rules.

## Before
Theree is no place in the UI to create the Automated membership upgrade rule nor the entity itself is created yet.

## After

A new entity is created called AutoMembershipUpgradeRule which contains the following main fields : 

- name : an auto generated machine name (generated from the label value that the user will input by replacing capital letters
to small letters and replacing spaces with underscores)
- label : the name of the new upgrade rule/
- from_membership_type_id : the membership type that we want this rule to apply on
- to_membership_type_id : the membership type that we want the "from membership type" to upgrade to.
- upgrade_trigger_date_type : which date should be used when calculating if it is the time to upgrade the membership, currently can be either the membership join date (member since) or the membership start date.
- period_length and period_length_unit : to how long we need the contact to have the "from membership" until it get upgraded. will be calculated in relative to the upgrade_trigger_date_type .
- filter_group : a group or smart group (both entities are stored inside civicrm_group table) to which we will limit which contacts group this rule will apply on.

the SQL to create this entity was added to this file `sql/autoupgraderuletable_install.sql` which will be executed automatically when we install the extension (since the file name ends with _install).  and also an upgrader (upgrade_0005) is added to create the table on existing sites by running the same SQL file.

also new form is created to allow users to create/update this new entity : 

2020-09-18 13_20_21-New Automated Membership Upgrade Rule _ me4test6

hence that currently this form is only accessible by hitting the following URL : 

**{site-url}/civicrm/admin/member/automated-upgrade-rules/add**

or by appending the id to the same URL in case of "update/edit"  : 

**{site-url}/civicrm/admin/member/automated-upgrade-rules/add&id={target-id}**

the menu item will be added later when we create the rules listing (management) page.

here is a gif for creating/updating the entity and how it will look using the API explorer : 

dadada


CRM/MembershipExtras/Test/Fabricator/MembershipType.php
```diff
@@ -9,7 +9,7 @@
 class CRM_MembershipExtras_Test_Fabricator_MembershipType extends BaseFabricator {
 
   /**
-   * Default parameters that will be used to fabricate a membership type.
+   * Default parameters that will be used to fabricate a membership type..
```
MiyaNoctem, line 12: Extra'.'?

MiyaNoctem (approved):

MiyaNoctem (dismissed):

PR #297, opened 2020-10-05
MAE-380:  Remove APIs and hook that will not be used for new UI

erawat:
## Overview

This PR is to remove Membership type APIs that were created for getting details membership types detail that used for calculating prorata and updating instalment. 

These codes are not currently used on the current version of membership extras and will not be used for the new UI on Membership Extra version 5. 

## Before

Below APIs were used on PaymentPlanToggler form but the code was commented out. 
- Getinstalmentamounts
- Getinstalmentamountsforpriceset

https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/master/templates/CRM/Member/Form/PaymentPlanToggler.tpl#L117

```
   if (instalmentFrequency == 1 && instalmentFrequencyUnit == 'month') {
          //Commented out for now till the BE functionality to store these values are in place.
          // updateInstalmentAmounts();
    }
```

Below APIs were used in prorata form hook. 

- Getproratedamountforpriceset.php
- Getdatesformembershiptype.php 
- Getproratedamount.php 

## After

- APIs will be deleted as they are no longer in used. 
- MembershipAmountProRata hook is also deleted as they will not be used. 

## Technical Details

Why delete this API instead of re-using them for the new UI?  This UI is actually returning values of membership type and the calculation is done on the front end which is currently incorrect. 

One of the requirements for the UI changes is to calculate the schedule / instalment on the backend and only use the front end to display the instalment. In order to achieve this, the new set of APIs will be created.


MiyaNoctem (approved):

erawat:
This PR was merged incorrectly to work stream branch. New PR will be created, work stream branch has been restored to the state before this PR is merged.

PR #305, opened 2020-10-28
MAE-380: Membership type rolling period instalment payment plan new screen

erawat:
## Overview

This PR is to simplify the payment plan on the "New Membership" screen.  

The instalments are generated by selected schedule or selected membership type. 

## Before

The existing instalment screen was removed and already merged to base branch so the instalment was not existed in this case. 

## After

When select membership type and/or schedule the instalment will be displayed based on schedule. 

Peek 2020-10-28 10-47

Currently, the new screen does not support  fixed membership type and price set. 

## Technical Details

This PR has refactored how instalments calculated from calculating the instalment on the font end (JavaScript) to calculate instalments  from the back end. 

What have been changed? 

- New CiviCRM API called 'PaymentSchedule' is created to handle. The API accepts four parameters which are membership_type_id, schedule, start_date, end_date, and join_date.  The API then will return the list of instalment as JSON as per below example with select quarterly schedule.

The calculation and format are done in the back end so the front end would only need to generate the display. 

`{"is_error":0,"version":3,"count":4,"values":[{"instalment_no":"1","instalment_date":"25th October 2020","instalment_tax_amount":"\u00a30","instalment_amount":"\u00a330","instalment_status":"Pending"},{"instalment_no":"2","instalment_date":"25th February 2021","instalment_tax_amount":"\u00a30","instalment_amount":"\u00a330","instalment_status":"Pending"},{"instalment_no":"3","instalment_date":"25th June 2021","instalment_tax_amount":"\u00a30","instalment_amount":"\u00a330","instalment_status":"Pending"},{"instalment_no":"4","instalment_date":"25th October 2021","instalment_tax_amount":"\u00a30","instalment_amount":"\u00a330","instalment_status":"Pending"}]}`

## Comment

There is one unit test fails but not related to this update and it will be fixed separately.


CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /**
+   * Calculates the remaining days from start date to the end of month date.
+   *
+   * @param DateTime $startDate
+   *
+   * @return int
+   */
+  private function calculateRemainingDaysInStartDateMonth(DateTime $startDate) {
```
MiyaNoctem, line 230: This method isn't used anywhere, right? If so, does it need to be deleted?

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+  }
+
+  /**
+   * Calculates the number of days in the month the start date falls in.
+   *
+   * @param DateTime $startDate
+   *
+   * @return int
+   */
+  private function getNumberOfDaysInStartDateMonth(DateTime $startDate) {
```
MiyaNoctem, line 244: Same for this method, looks like it isn't used anywhere.

api/v3/PaymentSchedule.php
```diff
@@ -0,0 +1,111 @@
...
+ */
+function civicrm_api3_payment_schedule_get($params) {
+  $joinDate = !empty($params['join_date']) ? new DateTime($params['join_date']) : NULL;
+  $startDate = !empty($params['start_date']) ? new DateTime($params['start_date']) : NULL;
+  $endDate = !empty($params['end_date']) ? new DateTime($params['end_date']) : NULL;
+
+  $membershipTypeID = $params['membership_type_id'];
+  $schedule = $params['schedule'];
+
+  if (!in_array($schedule, ['monthly', 'quarterly', 'annually'])) {
```
MiyaNoctem, line 62: Maybe:
```php
if (!in_array($schedule, [
  CRM_MembershipExtras_Service_MembershipTypeInstalmentCalculator::MONTHLY,
  CRM_MembershipExtras_Service_MembershipTypeInstalmentCalculator::QUARTERLY,
  CRM_MembershipExtras_Service_MembershipTypeInstalmentCalculator::ANNUALLY,
])) { 
  ... 
}
```

To have a single source of truth?

api/v3/PaymentSchedule.php
```diff
@@ -0,0 +1,111 @@
...
+  $spec['join_date'] = [
+    'name' => 'join_date',
+    'title' => 'Member Since Date',
+    'type' => CRM_Utils_Type::T_DATE,
+    'api.required' => 0,
+  ];
+}
+
+/**
+ * PaymentScheduleCalculation.get API
```
MiyaNoctem, line 46: PaymentSchedule.get API?

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+use CRM_MembershipExtras_Service_MoneyUtilities as MoneyUtilities;
+use CRM_MembershipExtras_Service_MembershipTypeTaxAmountCalculator as MembershipTypeTaxAmountCalculator;
+use CRM_MembershipExtras_Exception_InvalidMembershipTypeInstalmentCalculator as InvalidMembershipTypeInstalmentCalculator;
+use CRM_MembershipExtras_Service_MembershipTypeDatesCalculator as MembershipTypeDatesCalculator;
+use CRM_MembershipExtras_DTO_ScheduleInstalmentAmount as ScheduleInstalmentAmount;
+
+/**
+ * Class CRM_MembershipExtras_Service_MembershipTypeInstalmentCalculator
+ */
+class CRM_MembershipExtras_Service_MembershipTypeInstalmentCalculator {
```
omarabuhussein, line 12: I would rename this class maybe to : 

CRM_MembershipExtras_Service_MembershipInstalmentsSchedule

and rename  **generateInstalments** to **generate**

then I would reorder the methods in the order they are called in starting with the public method.

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+
+  const MONTHLY = 'monthly';
+  const QUARTERLY = 'quarterly';
+  const ANNUALLY = 'annually';
+
+  const MONTHLY_INSTALMENT = 12;
```
omarabuhussein, line 18: I would rename these 3 constants to end with _COUNT .. example : MONTHLY_INSTALMENT_COUNT

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+  const QUARTERLY_INSTALMENT = 4;
+  const ANNUALLY_INSTALMENT = 1;
+
+  /**
+   * @var \CRM_Member_BAO_MembershipTypeDatesCalculator
```
omarabuhussein, line 23: you are using wrong type hinting here, it should be CRM_MembershipExtras_Service_MembershipTypeDatesCalculator

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+   */
+  private $membershipTypeDatesCalculator;
+
+  /**
+   * @var array
+   */
+  private $membershipTypes;
+
+  /**
+   * @var \CRM_Member_BAO_MembershipTypeTaxAmountCalculator
```
omarabuhussein, line 33: same here, wrong type hinting

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+   * @return CRM_MembershipExtras_DTO_ScheduleInstalmentAmount
+   */
+  private function calculateInstalmentAmount() {
+    $totalMembershipTypeAmount = 0;
+    $totalMembershipTypeTaxAmount = 0;
+    foreach ($this->membershipTypes as $membershipType) {
+      $totalMembershipTypeAmount += $membershipType->minimum_fee;
+      $totalMembershipTypeTaxAmount = +$this->membershipTypeTaxAmountCalculator->calculateTax($membershipType);
+    }
+    switch ($this->schedule) {
```
omarabuhussein, line 78: I would add a space before and after the switch

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
```diff
@@ -0,0 +1,248 @@
...
+  /**
+   * Calculates the first instalment amount for a set of membership types given that the conditions
+   * for calculating the first instalment amount is met.
+   *
+   * @param CRM_MembershipExtras_DTO_ScheduleInstalmentAmount $instalmentAmount
+   * @param \DateTime|NULL $startDate
+   * @param \DateTime|NULL $endDate
+   * @return CRM_MembershipExtras_DTO_ScheduleInstalment
+   */
+  private function calculateFirstInstalment(
```
omarabuhussein, line 109: I don't really understand the point of having this method, I don't see it doing anything special, can you elaborate more on why is it needed?

omarabuhussein (changes requested):
So I left some notes but I also have some other generic  notes/questions : 

1- The commit messages and their order and what each commit contain is really a mess, the first commit, for example, contains a lot of debugging and commented out code, can you maybe reorganize them more properly, interactive rebase can help with that maybe or whatever method you prefer, also commit messages like "MAE-380: Rename classes to reflect what they do" or "MAE-380: Format the function" does not reflect the actual purpose of your PR. 

2- What about fixed membership types and price sets, are you planning to start working on them soon, I am saying that because they already part of the ticket scope (MAE-380), I am fine if you are doing them in another PR (or more preferably creating a new ticket for them) as long as they are not forgotten.

3- I am not sure if we decided to support quarterly payment plans or not, maybe something to confirm with Ana since she was planning to check on that.

4- Not sure if this is part the specs documentation, but what about "manual" renewal screen does this work with it? maybe to be checked with Ana too.

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
erawat: @MiyaNoctem  I left it there for using for calculating fixed period type.

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
erawat: @MiyaNoctem as describe above, I left it there for using for calculating fixed period type.

CRM/MembershipExtras/Service/MembershipTypeInstalmentCalculator.php
erawat: @omarabuhussein  we would need to calculate first instalment different for fixed period membership type.

CRM/MembershipExtras/Service/MembershipTypeDatesCalculator.php
```diff
@@ -7,22 +7,28 @@ class CRM_MembershipExtras_Service_MembershipTypeDatesCalculator {
...
+   * @param CRM_Member_BAO_MembershipType $membershipType
    * @param \DateTime|NULL $startDate
    * @param \DateTime|NULL $endDate
    * @param \DateTime|NULL $joinDate
    *
    * @return array
    */
-  public function getDatesForMembershipType(MembershipType $membershipType, DateTime $startDate = NULL, DateTime $endDate = NULL, DateTime $joinDate = NULL) {
+  public function getDatesForMembershipType(
+    MembershipType $membershipType,
```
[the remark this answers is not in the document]
erawat: It was implemented as it is, will update it.

CRM/MembershipExtras/DTO/ScheduleInstalmentAmount.php
```diff
@@ -0,0 +1,37 @@
+<?php
+
+
+class CRM_MembershipExtras_DTO_ScheduleInstalmentAmount {
```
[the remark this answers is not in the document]
erawat: @omarabuhussein  they are actually different objects.

erawat:
> So I left some notes but I also have some other generic notes/questions :
> 
> 1- The commit messages and their order and what each commit contain is really a mess, the first commit, for example, contains a lot of debugging and commented out code, can you maybe reorganize them more properly, interactive rebase can help with that maybe or whatever method you prefer, also commit messages like "MAE-380: Rename classes to reflect what they do" or "MAE-380: Format the function" does not reflect the actual purpose of your PR.
> 
Commits will be squashed to make it cleaner. 

> 2- What about fixed membership types and price sets, are you planning to start working on them soon, I am saying that because they already part of the ticket scope (MAE-380), I am fine if you are doing them in another PR (or more preferably creating a new ticket for them) as long as they are not forgotten.
> 

This will be done in another PR, perhaps with sub ticket for this as I want this PR to be reviewed, so we could agree on the approach before implementing fixed period membership type. 

> 3- I am not sure if we decided to support quarterly payment plans or not, maybe something to confirm with Ana since she was planning to check on that.
> 

I won't support quarterly payment for fixed membership type but for rolling membership type, we are supporting them. 

> 4- Not sure if this is part the specs documentation, but what about "manual" renewal screen does this work with it? maybe to be checked with Ana too.

Will check.

api/v3/PaymentSchedule.php
```diff
@@ -0,0 +1,111 @@
...
+ *
+ * @param array $params
+ * @return array API result descriptor
+ * @throws API_Exception
+ * @throws CRM_MembershipExtras_Exception_InvalidMembershipTypeInstalmentAmount|CiviCRM_API3_Exception
+ * @throws Exception
+ */
+function civicrm_api3_payment_schedule_get($params) {
```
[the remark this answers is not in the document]
erawat: @omarabuhussein  I don't mind changing the API name, and the service class. However, on the point 2, I totally agree that we may need to come up with better structure for the extension. However, on the current implementation of this API. The code after line 68 is technically belonging to the API, what it does is to calculate dates for each membership type MembershipTypeDatesCalculator then call MembershipInstalmentsSchedule to generate instalment and finally we format arrays and objects return from generate method and pass the result to civicrm API. 

And for your question for reusing it again in another part inside the extension, we would need to call MembershipInstalmentsSchedule instead of the API.

CRM/MembershipExtras/DTO/ScheduleInstalmentAmount.php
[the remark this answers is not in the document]
omarabuhussein: > we would need to use array for storing instalment date and instalment amount inside the Schedule instalment class

Can you explain a bit what do you mean by this, better with little code example ? because I don't understand why you would need to use an array inside ScheduleInstalment class if you are going to decompose these two DTOs.

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
```diff
@@ -0,0 +1,183 @@
...
+   * @return mixed
+   * @throws Exception
+   */
+  private function getRollingMembershipInstalments($membershipTypes, $schedule) {
+    $membershipTypeInstalmentCalculator = $this->getMembershipTypeInstalmentAmount($membershipTypes, $schedule);
+
+    $startDate = new DateTime();
+    $joinDate = $startDate;
+    $endDate = new DateTime();
+    $endDate->add(new DateInterval('P1Y'));
```
omarabuhussein, line 134: the end date should be after X amount of time (as defined by the membership type) minus 1 day

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
```diff
@@ -0,0 +1,183 @@
...
+
+    $this->assertCount(MembershipInstalmentsSchedule::MONTHLY_INSTALMENT_COUNT, $instalments);
+
+    $expectedAmount = MoneyUtilities::roundToPrecision(
+      ($membershipTypes[0]->minimum_fee + $membershipTypes[1]->minimum_fee) / MembershipInstalmentsSchedule::MONTHLY_INSTALMENT_COUNT,
+      2
+    );
+
+    $instalment = $instalments[0];
+    $this->assertEquals($expectedAmount, $instalment->getInstalmentAmount()->getAmount());
```
omarabuhussein, line 35: you will need to also test that each instalment has the correct date, and I would do it in separate test method, and I will rename this one so you end up with : 

- testRollingMonthlyInstalmentsScheduleAmounts
- testRollingMonthlyInstalmentsScheduleDates

and the same goes for annually and quarterly

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  /**
+   * Test Generate Rolling Annually Instalment
+   *
+   * @throws Exception
+   */
+  public function testCalculateRollingAnnuallyInstalments() {
```
omarabuhussein, line 66: you used `generate` in the above test methods and this should follow that, but anyway in my comment above I suggested to just remove it.

tests/phpunit/api/v3/PaymentScheduleTest.php
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * Class PaymentScheduleTest
+ */
+class PaymentScheduleTest extends BaseHeadlessTest {
+
+  /**
+   * @throws CiviCRM_API3_Exception
+   */
+  public function testGetMonthlyRollingMembershipTypeSchedule() {
```
omarabuhussein, line 13: I would focus my test on the stuff that the API itself is doing, which include : 

1- testing something other than these : 

```

    CRM_MembershipExtras_Service_MembershipInstalmentsSchedule::MONTHLY,
    CRM_MembershipExtras_Service_MembershipInstalmentsSchedule::QUARTERLY,
    CRM_MembershipExtras_Service_MembershipInstalmentsSchedule::ANNUALLY,
```

will result in throwing an exception

2- test the format of the output, for example, that it should be an array with X elements (which vary if it monthly or annual or ..etc). as well as if at least one instalment in the array contains the following keys : 
`instalment_no`, `instalment_date`, 'instalment_tax_amount', 'instalment_amount' ..etc. And hence not test the content but rather if these keys exsit.

omarabuhussein (changes requested):

CRM/MembershipExtras/DTO/ScheduleInstalmentAmount.php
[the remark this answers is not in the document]
erawat: @omarabuhussein  what I meant, was not actually to have an array inside ScheduleInstalment but we would need to use the array for instalment amount, so we might store instalment amount as an array too. 

So why I don't want to decompose these DTOs.  The ScheduleInstalmentAmount class is used as an object for storing the instalment amount that we calculate and it will be passed to calucateFirstInstalment method for getting first instalment amount (depending on membership type / period). See below

https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/MAE-380-membership-type-rolling-period-instalment/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php#L158

https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/MAE-380-membership-type-rolling-period-instalment/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php#L175

So the actual schedule instalment https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/MAE-380-membership-type-rolling-period-instalment/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php#L100

So if we are to decompose it, we might need to change or rename below function from calculateInstalmentAmount() to something like calculateInstalment() or getInstalment() instead?

https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/MAE-380-membership-type-rolling-period-instalment/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php#L136

But at the end, we would need ScheduleInstalment DTO would need to re-initiate when we generate an array that contain a list of different instalment objects here. https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/MAE-380-membership-type-rolling-period-instalment/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php#L100

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
erawat: @omarabuhussein  currently, we support monthly for 1 year  rolling membership type at this stage, I will implement / refactor when we are implementing, quarterly, and annually.

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
erawat: @omarabuhussein  noted on this, I'll leave annually and quarterly to the next PR as this PR is only related to monthly instalment only.

tests/phpunit/api/v3/PaymentScheduleTest.php
erawat: @omarabuhussein   
Point 1 is done, expected exception added. 
Point 2: I may disagree with this, the reason I would check the content because the API is also return with format for example, currency symbol, so I would like to assert if they are equals. If the fields does not exist, the test will fail anyway.

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
omarabuhussein: but the end date of the membership is related to its duration and not if it paid annually or monthly or etc, and it always be `-1 day` from the (start date + membership duration)

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
omarabuhussein: this again has nothing to do with it being annually or quarterly, actually, it even makes more sense to be done for monthly that anything else since you need to check that the 12 instalments dates are correct.

omarabuhussein (approved):
I am approving this knowing that "GitHub actions" is unable to create sites currently, this will be fixed soon on the Master branch and then @erawat  will rebase the workstream branch with master, and then we can run the run the actions again and confirm that there are no broken tests.

PR #312, opened 2020-11-16
MAE-397: Fix Membership Deletion

MiyaNoctem:
## Overview
When you try to delete a membership, a popup is shown to confirm the deletion. After confirming, popup keeps on running on the same page and no deletion is happens.

## Before
As part of the membership deletion process, all contributions related to the membership are deleted to. After deletion, a post hook is triggered, which in turn executes our Post membership payment hook. This hook used a getvalue API call on the Contribution's entity to obtain the related recurring contribution, however, by this time related contributions had already been deleted. Thus an exception was thrown by the API call.

## After
Changed the API call from getvalue to get, so we can check the count before trying to access the value for recurring contribution ID.

### DIFF
--- a/CRM/MembershipExtras/Hook/Post/MembershipPayment.php
+++ b/CRM/MembershipExtras/Hook/Post/MembershipPayment.php
@@ -51,10 +51,18 @@ class CRM_MembershipExtras_Hook_Post_MembershipPayment {
   }
 
   private function setRelatedRecurContributionId() {
-    $this->relatedRecurContributionId = civicrm_api3('Contribution', 'getvalue', [
+    $this->relatedRecurContributionId = NULL;
+
+    $result = civicrm_api3('Contribution', 'get', [
+      'sequential' => 1,
       'id' => $this->membershipPayment->contribution_id,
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Hook/Post/MembershipPayment.php
```diff
@@ -51,10 +51,15 @@ public function __construct($operation, $objectId, CRM_Member_DAO_MembershipPaym
   }
 
   private function setRelatedRecurContributionId() {
-    $this->relatedRecurContributionId = civicrm_api3('Contribution', 'getvalue', [
-      'id' => $this->membershipPayment->contribution_id,
-      'return' => 'contribution_recur_id',
-    ]);
+    try {
+      $this->relatedRecurContributionId = civicrm_api3('Contribution', 'getvalue', [
```
erawat, line 55: @MiyaNoctem  as discussed, perhaps we should call API and check count instead of using try {} catch here

erawat (changes requested):

erawat (approved):

PR #314, opened 2020-11-17
MAE-400: Fix Receive Date on Fixed Single Installment Renewal

MiyaNoctem:
## Overview
If you create a new fixed membership to a contact with Auto-renew set, payment plan 1 installment and with the 
following dates:
- Member Since - 1 July 2019
- Member Start - 1 July 2019
- Member End Date - 30/11/2019 (Auto-filled)
- Payment Received Date - 1 July 2019

After renewing, the dates will get updated to:
- Member Since - 1 July 2019 - Correct
- Member Start - 1 July 2019  - Correct
- Member End Date - 30/11/2020 - Correct
- New Contribution Date - 1 July 2019 - Incorrect

New contribution date should be 1 Dec 2019

## Before
When renewing a payment plan with a single installment, the new recieve date is calculated from the initial cycle day of the payment plan, which was the 1st of july (day 182 of the year). Thus, all contribution recieve dates are calculated to be on that day of the corresponding year.

## After
Fixed by setting the contribution's receive date as the memberships' calculated start date if there is at least one fixed membership in the payment plan. For the evaluated scenario, this would mean the recieve date of the new contribution would be 1st of Dec, 2019.

## Notes
Added a test to verify the scenario.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/SingleInstallmentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/SingleInstallmentPlan.php
@@ -123,7 +123,11 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_SingleInstallmentPlan extends 
    */
   public function renew() {
     $this->membershipsStartDate = $this->calculateRenewedMembershipsStartDate();
-    $this->paymentPlanStartDate = $this->calculateNoInstallmentsPaymentPlanStartDate();
+
+    $this->paymentPlanStartDate = $this->membershipsStartDate;
+    if (!$this->areAnyMembershipsFixed()) {
+      $this->paymentPlanStartDate = $this->calculateNoInstallmentsPaymentPlanStartDate();
+    }
 
     $this->endCurrentLineItemsAndCreateNewOnesForNextPeriod($this->currentRecurContributionID);
     $this->updateRecurringContributionAmount($this->currentRecurContributionID);
@@ -133,6 +137,28 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_SingleInstallmentPlan extends 
 
     $this->recordPaymentPlanFirstContribution();
     $this->renewPaymentPlanMemberships($this->currentRecurContributionID);
+  }
+
+  /**
+   * Checks if any of the memberships in the plan are fixed.
+   *
+   * @return bool
+   */
+  private function areAnyMembershipsFixed() {
+    $currentPeriodLines = $this->getRecurringContributionLineItemsToBeRenewed($this->currentRecurContributionID);
+
+    foreach ($currentPeriodLines as $lineItem) {
+      if ($lineItem['entity_table'] != 'civicrm_membership') {
+        continue;
+      }
+
+      if ($lineItem['period_type'] === 'fixed') {
+        return TRUE;
+      }
+
+    }
+
+    return FALSE;
   }
 
   /**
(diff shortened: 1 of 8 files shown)


jamienovick:
@omarabuhussein @ana-compucorp  Can you take a look at this? I'm not sure I agree with the logic here?

We are back at this question on how the next contribution should be calculated? I thought we agreed it should be based on the next contribution date field?

Also @MiyaNoctem  - this logic seems odd: 

"Fixed by leaving membership start date as contribution's receive date if there is at least one fixed membership in the payment plan."

That doesnt appear to be what is changed above?

MiyaNoctem:
@jamienovick , I've changed the wording forthe "After" part of the PR so it's clearer, but the fix I did is to set the recieve date of the contribution to the membership's start date. In the scenario we're evaluating, this would mean the new contribution would have 2019-12-01 as receive date.

We discussed the issue this morning in stand-up, and also though about how the system should behave if there are both fixed and rolling memberships in the same plan, and if it is even a valid use case. We will block this ticket until we better define the expected behaviour.

omarabuhussein:
@james-compucorp 

> Can you take a look at this? I'm not sure I agree with the logic here?
We are back at this question on how the next contribution should be calculated? I thought we agreed it should be based on the next contribution date field?

We did not start the work on this stuff yet , so this PR is more about fixing an issue with the current behaviour in which we can revisit after working on the new changes.


@MiyaNoctem 

> We discussed the issue this morning in stand-up, and also about how the system should behave if there are both fixed and rolling memberships in the same plan and if it is even a valid use case. We will block this ticket until we better define the expected behaviour.

so I was thinking, and probably the new contribution (After autorenewal) date should be the least "new" start date of the memberships in the payment plan.

CRM/MembershipExtras/Test/Fabricator/Base.php
```diff
@@ -33,10 +33,29 @@ public static function fabricate(array $params = []) {
       throw new \Exception('Entity name cannot be empty!');
     }
 
-    $params = array_merge(static::$defaultParams, $params);
-    $result = civicrm_api3(static::$entityName, 'create', $params);
+    $defaultParams = static::getDefaultParameters();
```
omarabuhussein, line 36: you will need to re-write your tests according to the latest changes in the master branch.

CRM/MembershipExtras/Job/OfflineAutoRenewal/SingleInstallmentPlan.php
```diff
@@ -123,7 +123,11 @@ protected function getRecurringContributions() {
    */
   public function renew() {
     $this->membershipsStartDate = $this->calculateRenewedMembershipsStartDate();
-    $this->paymentPlanStartDate = $this->calculateNoInstallmentsPaymentPlanStartDate();
+
+    $this->paymentPlanStartDate = $this->membershipsStartDate;
+    if (!$this->areAnyMembershipsFixed()) {
+      $this->paymentPlanStartDate = $this->calculateNoInstallmentsPaymentPlanStartDate();
```
omarabuhussein, line 129: so usually the new payment plan start date should be same as the new minimum membership start date in the payment plan, and calculateNoInstallmentsPaymentPlanStartDate() here just rectify the date to be aligned with the payment plan cycle date. But this is also  for single instalment (annual) payment plans, and we are planning soon to get rid of using cycle day for annual payment plans, so probably this will even be changed later to just this : 

```
$this->paymentPlanStartDate = $this->membershipsStartDate;
```

so I think I am fine with this solution for the time being.

omarabuhussein (changes requested):

MiyaNoctem:
Closed in favor of #321

PR #315, opened 2020-11-19
MAE-401: Fix Renewal of Plans with Only New Membership for Next Period

MiyaNoctem:
## Overview
If you create a renewable payment plan with one membership and multiple instalments, and before renewing you set the membership not to auto-renew, and add a new membership for the next period, once auto-renew runs the amount for the new recurring contribution will be 0, and if you enter the manage future instalments screen, the plan will have no line items for the current period.

## Before
When renewing, the start date of the new lines was being calculated by checking the latest end date of all the renewable memberships in the plan and adding a day. Since there are no memberships to be renewed, the start date for the lines was left empty, which in turn caused:
- the amount for the recurring contribution was set to 0, as the calculation only checked lines with start date, 
- the manage instalments screen showed no lines, as it only shows lines with a start date.

## After
Fixed by checking all lines associated to a payment plan, regardless of their renewable status, so that it takes the latest end date of renewable memberships, but if there are no renewable memberships, it takes the latest end date of non-renewable memberships.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
@@ -269,7 +269,7 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
   /**
    * Obtains the list of recurring line items to be renewed for the plan.
    *
-   * Returns an array with all lthe line items of the payment plan that are not
+   * Returns an array with all the line items of the payment plan that are not
    * removed and are set to auto-renew.
    *
    * @param int $recurringContributionID
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
@@ -267,14 +267,28 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
   }
 
   /**
-   * Obtains the list of recurring line items to be renewed for the payment plan
-   * being renewed.
+   * Obtains the list of recurring line items to be renewed for the plan.
+   *
+   * Returns an array with all lthe line items of the payment plan that are not
+   * removed and are set to auto-renew.
    *
    * @param int $recurringContributionID
    *
    * @return array
    */
   abstract protected function getRecurringContributionLineItemsToBeRenewed($recurringContributionID);
+
+  /**
+   * Obtains list of all active line items of the given recurring contribution.
+   *
+   * Returns an array with all the line items that ar not removed from the
+   * payment plan, irrespective if they are renewable or not.
+   *
+   * @param int $recurringContributionID
+   *
+   * @return array
+   */
+  abstract protected function getAllRecurringContributionActiveLineItems($recurringContributionID);
 
   /**
    * Obtains list of recurring line items that are active for the new recurring
@@ -611,7 +625,7 @@ abstract class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentPlan {
     $membershipCreateResult = civicrm_api3('Membership', 'create', [
       'contact_id' => $this->currentRecurringContribution['contact_id'],
       'membership_type_id' => $priceFieldValue['membership_type_id'],
-      'join_date' => date('YmdHis'),
+      'join_date' => $this->membershipsStartDate,
       'start_date' => $this->membershipsStartDate,
       'end_date' => $lineItem['end_date'],
       'contribution_recur_id' => $this->newRecurringContributionID,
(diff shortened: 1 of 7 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
```diff
@@ -880,17 +897,24 @@ protected function calculateRenewedMembershipsStartDate() {
       }
 
       $membershipEndDate = new DateTime($lineItem['memberhsip_end_date']);
-      if (!isset($latestDate)) {
-        $latestDate = $membershipEndDate;
+      $isLineAutoRenewable = $lineItem['auto_renew'] == '1' ? 'auto_renewable' : 'not_auto_renewable';
+
+      if (!isset($latestEndDates[$isLineAutoRenewable])) {
```
omarabuhussein, line 902: what if you have two memberships in the current tab that we set to "not" renew, and added only one to the next period tab, I think it make sense to pick the latest end date in such case too, but you are not doing and date comparison for that, so I think it make more sense to change these lines to :

```
$membershipEndDate = new DateTime($lineItem['memberhsip_end_date']);

if (empty($lineItem['auto_renew']) && $latestEndDates['not_auto_renewable'] < $membershipEndDate) {
$latestEndDates['not_auto_renewable'] = $membershipEndDate;
}


if (!empty($lineItem['auto_renew']) && $latestEndDates['auto_renew'] < $membershipEndDate) {
$latestEndDates['auto_renew'] = $membershipEndDate;
}

```

tests/phpunit/CRM/MembershipExtras/Job/OfflineAutoRenewal/MultiInstallmentPlanTest.php
```diff
@@ -761,4 +763,98 @@ private function getCustomFieldId($customGroupName, $customFieldName) {
...
+      'name' => 'Test New Rolling Membership',
+      'period_type' => 'rolling',
+      'minimum_fee' => 1200,
+      'duration_interval' => 1,
+      'duration_unit' => 'year',
+    ]);
+
+    $multipleInstallmentRenewal = new MultipleInstallmentRenewalJob();
+    $multipleInstallmentRenewal->run();
+    $this->isPaymentPlanMembershipRenewed($paymentPlan['id'], '+1 year -1 month -1 day');
```
omarabuhussein, line 794: this method only returns true or false, so calling it like that does not test for anything and instead you should assert the result of this method (see the other tests that call it), but I think the asserts below should be enough so I think it is fine to remove it complety.

tests/phpunit/CRM/MembershipExtras/Job/OfflineAutoRenewal/MultiInstallmentPlanTest.php
```diff
@@ -761,4 +763,98 @@ private function getCustomFieldId($customGroupName, $customFieldName) {
     ]);
   }
 
+  public function testRenewalWithNonRenewableLineOnCurrentPeriodAndNewMembershipForNextPeriod() {
```
omarabuhussein, line 766: I would also add the following tests : 

1- Two memberships in the current period where non of them is renewal and each have different end date, with one renewal membership in the next tab with.

2- Two memberships in the current period where one of them is renewal and the other is not and each with  diffrenet end date, and another renewal membership in the next tab (which one membership will be renewed and new one with new type will be created).

also, similar tests for single instalment job should be added.

omarabuhussein (changes requested):

CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentPlan.php
MiyaNoctem: I think this is equivalent to this part:
```
      if (!isset($latestEndDates[$isLineAutoRenewable])) {
        ...
      elseif ($latestEndDates[$isLineAutoRenewable] < $membershipEndDate) {
        $latestEndDates[$isLineAutoRenewable] = $membershipEndDate;
      }
```

I mean, if `$lineItem['auto_renew']` is empty, then my code translates to:
```
      if (!isset($latestEndDates['not_auto_renewable'])) {
        ...
      elseif ($latestEndDates['not_auto_renewable'] < $membershipEndDate) {
        $latestEndDates['not_auto_renewable'] = $membershipEndDate;
      }
```

Which is equivalent to what you wrote for the empty autor_renew flag scenario, without the `empty($lineItem['auto_renew'])`, which I don't need, as I already know it is empty. It just factorizes both cases in the same code.

In the scenario you describe, with the two memberships for current period that are not auto-renew, plus a new membership on the next period, the system will still pick the latest end date. For example:

- Membership 1, not auto-renew, ends on 2020-06-30
- Membership 2, not auto-renew, ends on 2020-12-31
- Membership 3 for new period.

In this case, the list of active memberships would be Membership 1 and Membership 2. At the end of the first cycle, the array with the end dates will be:
```
[
  'not_auto_renewable' => '2020-06-30',
  'auto_renew' => null,
]
```

At the end of the second cycle:
```
[
  'not_auto_renewable' => '2020-12-31',
  'auto_renew' => null,
]
```

Finally, since the latest end date for auto renewable memberships is null, it uses 2020-12-31, which is correct.

omarabuhussein (approved):

PR #316, opened 2020-11-25
MAE-283: Prevent Membership Extension on Bulk Contribution Update

MiyaNoctem:
## Overview
Membership end date is being updated to next period after doing contribution bulk action 'Updating Pending Contribution Status'.

## Before
When updating contributions in bulk, CiviCRM will set the num_terms parameter which, in their own words, is used as a "special sauce" to cause the memberships related to the contributions to be extended. Furthermore, this extension is calculated in the membership API call, before hitting the BAO.

https://github.com/civicrm/civicrm-core/blob/master/api/v3/Membership.php#L106-L108

## After
Neutralized CiviCRM API's special sauce, by implementing a pre hook on the membership entity that will unset the end date if it detects a bulk contribution status update is being done.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
@@ -104,11 +104,11 @@ class CRM_MembershipExtras_Hook_Pre_MembershipEdit {
    * @throws \CRM_Core_Exception
    */
   private function isBulkStatusUpdate() {
-    $q = CRM_Utils_Request::retrieve('q', 'String');
     $statusNext = CRM_Utils_Request::retrieve('_qf_Status_next', 'String');
     $contributionStatusID = CRM_Utils_Request::retrieve('contribution_status_id', 'String');
+    $currentPath = CRM_Utils_System::currentPath();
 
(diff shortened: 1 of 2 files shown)


CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
```diff
@@ -176,10 +194,70 @@ private function isOfflinePaymentPlanContribution($recurringContributionID) {
...
+   * @return mixed|null
+   * @throws \CiviCRM_API3_Exception
+   */
+  private function getRecurringContributionFromMembership($membershipID) {
+    $membership = civicrm_api3('Membership', 'getsingle', [
+      'sequential' => 1,
+      'id' => $membershipID,
+      'options' => ['limit' => 0],
+    ]);
+    if (isset($membership['contribution_recur_id']) && !empty($membership['contribution_recur_id'])) {
```
omarabuhussein, line 219: `!empty` without `isset` should be enough

omarabuhussein (changes requested):
can you also write tests for this class CRM_MembershipExtras_Hook_Pre_MembershipEdit  ?

omarabuhussein (dismissed):

omarabuhussein (dismissed):

omarabuhussein (approved):

PR #323, opened 2020-12-07
IO-124: Test Cycle Day is Calculated From Receive Date

MiyaNoctem:
## Overview
We need to be sure the cycle day for monthly instalments is assigned according to the Next Payment Collection Run dates available.

For this, a hook was added in #322 , that can be implemented by Direct Debit extension to change the receive date so it matches the next available Payment Collection Run date. In turn we need to make sure cycle day also corresponds to the next available payment collection date.

## Before
There were no tests verifying cycle day is calculated from the first instalment's receive date.

## After
Added tests to make sure cycle day is calculated from first instalment's receive date.


MiyaNoctem:
Closing due to extra commits that shouldn¿t be part of this PR.

PR #325, opened 2020-12-16
IO-124: Add Tests to Check Cycle Day is Calculated From Receive Date

MiyaNoctem:
## Overview
We need to be sure the cycle day for monthly instalments is assigned according to the Next Payment Collection Run dates available.

For this, a hook was added in #322 , that can be implemented by Direct Debit extension to change the receive date so it matches the next available Payment Collection Run date. In turn we need to make sure cycle day also corresponds to the next available payment collection date.

## Before
There were no tests verifying cycle day is calculated from the first instalment's receive date.

## After
Added tests to make sure cycle day is calculated from first instalment's receive date.

### DIFF
--- a/tests/phpunit/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessorTest.php
+++ b/tests/phpunit/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessorTest.php
@@ -18,7 +18,7 @@ class CRM_MembershpExtras_Hook_Pre_MembershipPaymentPlanProcessorTest extends Ba
       'is_pay_later' => TRUE,
       'skipLineItem' => 1,
       'skipCleanMoney' => TRUE,
-      'receive_date' => '2020-06-27',
+      'receive_date' => date('Y-m-27'),
       'contact_id' => $contact['id'],
       'fee_amount' => 0,
       'net_amount' => "1200",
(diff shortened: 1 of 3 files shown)


tests/phpunit/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessorTest.php
```diff
@@ -0,0 +1,147 @@
...
+    $_REQUEST['installments'] = 12;
+    $_REQUEST['installments_frequency'] = 1;
+    $_REQUEST['installments_frequency_unit'] = 'month';
+
+    $contact = ContactFabricator::fabricate();
+    $params = [
+      'is_pay_later' => TRUE,
+      'skipLineItem' => 1,
+      'skipCleanMoney' => TRUE,
+      'receive_date' => '2020-01-01',
```
erawat, line 89: @MiyaNoctem  would be good if we can have dates in the test dynamic instead of specific date as the test would be awkward if we re-visit it in 2022?.

erawat: @madeleinevironda  this would be for dates in line 21, 100 and 111

erawat (approved):

PR #326, opened 2020-12-23
IO-122: Add API to support price field values option

erawat:
## Overview

This PR adds / refactors existing Payment Schedule API to support getByPriceFieldValues and getByMembershipType actions. 

This PR also includes new fabricators, refactoring code and add more testes to the instalment schedule classes as well adding API helper that shares usage by both actions.

## Before

Payment Schedule API only supports membership type as parameters

## After

Payment Schedule API supports membership type and price field values through different actions. 
 
## Technical Details

There are details that worth mentioning on this section. 

**APIs**

_PaymentSchedule.getByPriceFieldValues_  - This API requires selected price field values which can be multiple, so the field price_field_values must be  that passed to the API as an array that includes selected quantity for non membership type price field value and quantity equals 1 if a membership type is attached to price field values.  Since we only support an array of price_field_values, an operator "IN" is the only operators that support for this API. 

_PaymentSchedule.getByMembershipTypeId_ - This API requires selected membership type ID

More details about how to use the API can be found in the API unit tests https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/IO-122-simplify-admin-ui/tests/phpunit/api/v3/PaymentScheduleTest.php

as well as the implementation on the PaymentPlanToggler 
https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/IO-122-simplify-admin-ui/templates/CRM/Member/Form/PaymentPlanToggler.tpl#L109

### Comments

This PR supports only rolling membership type as the fixed membership type as prorata will be implemented in the next stage.


CRM/MembershipExtras/Helper/PaymentScheduleAPI.php
```diff
@@ -0,0 +1,92 @@
...
+   * Gets instalment from the API params and membership types
+   *
+   * @param array $params
+   * @param array $membershipTypes
+   * @param array $nonMembershipPriceFieldValue
+   * @return array
+   * @throws CRM_MembershipExtras_Exception_InvalidMembershipTypeInstalmentCalculator
+   * @throws Exception
+   */
+  public function getInstalments(array $params, array $membershipTypes, array $nonMembershipPriceFieldValue = []): array {
```
omarabuhussein, line 32: can you please get rid of all return type declaration in this PR, my reasons are : 

- you are doing it in some places but leaving others in this PR
- we did not communicate together as a team if we want to start doing that.
- we also use @return in docblocks, I know it is not exactly the same but it still duplicates information.

 I just don't see it as a right step to be done at this point, and there are other important improvements we need to work on before looking at things like this and other stuff like static typing..etc

CRM/MembershipExtras/Test/Fabricator/FinancialAccount.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+use CRM_MembershipExtras_Test_Fabricator_Base as BaseFabricator;
+
+class CRM_MembershipExtras_Test_Fabricator_FinancialAccount extends BaseFabricator {
+
+  protected static $entityName = 'FinancialAccount';
+
+  protected static $defaultParams = [
```
omarabuhussein, line 9: these are not defaults, these are very specific values for a very specific scenario, I would rather keep them empty and set their values in the test classes that use this fabricator, same in EntityFinancialAccount

CRM/MembershipExtras/Test/Fabricator/FinancialType.php
```diff
@@ -0,0 +1,14 @@
...
+
+use CRM_MembershipExtras_Test_Fabricator_Base as BaseFabricator;
+
+class CRM_MembershipExtras_Test_Fabricator_FinancialType extends BaseFabricator {
+
+  protected static $entityName = 'FinancialType';
+
+  protected static $defaultParams = [
+    'sequential' => 1,
+    'name' => "Member Dues",
```
omarabuhussein, line 11: Member dues also should not be a default param

omarabuhussein (changes requested):

tests/phpunit/CRM/MembershipExtras/API/PaymentSchedule/MembershipTypeTest.php
```diff
@@ -0,0 +1,66 @@
...
+ * Class CRM_MembershipExtras_API_PaymentSchedule_MembershipTypeTest
+ *
+ * @group headless
+ */
+class CRM_MembershipExtras_API_PaymentSchedule_MembershipTypeTest extends BaseHeadlessTest {
+
+  /**
+   * @throws API_Exception
+   */
+  public function testValidateSchedule() {
```
omarabuhussein, line 15: I would rename this to : 

testInvalideScheduleWillThrowAnException

omarabuhussein (changes requested):

omarabuhussein (approved):

PR #331, opened 2021-01-07
MAE-447: Calculate Start Date From Previous End Date on Manual Renewal

MiyaNoctem:
## Overview
When a membership is manually renewed and changed to a payment plan, the membership end date is increased by a year as expected. But manage instalment start date is using previous year start date still.

## Before
The start date used on recurring line items was the original start date of the membership, which did not change as CiviCRM core only changes it if there is a gap in the membership's history.

## After
Used the membership's end date to calculate the projected start date by subtracting the appropriate number of periods as per the membership type's configuration.

### DIFF
--- a/CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
+++ b/CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
@@ -9,15 +9,15 @@ class CRM_MembershipExtras_Service_MembershipEndDateCalculator {
    * Calculates the membership new end date
    * for renewal.
    *
-   * @param int $membershipId
+   * @param int $membershipID
    *
    * @return string
    * @throws \CiviCRM_API3_Exception
    * @throws \CiviCRM_API3_Exception|\Exception
    */
-  public static function calculate($membershipId) {
+  public static function calculate($membershipID) {
     $newEndDate = 'null';
-    $membershipDetails = self::getMembership($membershipId);
+    $membershipDetails = self::getMembership($membershipID);
     $interval = self::getMembershipPeriodInterval($membershipDetails);
     if (!empty($interval)) {
       $currentEndDate = new DateTime($membershipDetails['end_date']);
@@ -40,7 +40,6 @@ class CRM_MembershipExtras_Service_MembershipEndDateCalculator {
     $previousEndDate = 'null';
     $membershipDetails = self::getMembership($membershipID);
     $interval = self::getMembershipPeriodInterval($membershipDetails);
-
     if (!empty($interval)) {
       $currentEndDate = new DateTime($membershipDetails['end_date']);
       $currentEndDate->sub(new DateInterval($interval));
--- a/CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
+++ b/CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
@@ -1,5 +1,8 @@
 <?php
 
+/**
+ * Class CRM_MembershipExtras_Service_MembershipEndDateCalculator.
+ */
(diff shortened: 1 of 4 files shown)


erawat:
@MiyaNoctem  I have a couple of notes before starting to review. 

1. Can you fix the failing linter? 
2. Since you are working on CRM_MembershipExtras_Hook_PostProcess_RecurringContributionLineItemCreator and CRM_MembershipExtras_Service_MembershipEndDateCalculator and you add unit tests for these class as  they are quite critical.

CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
```diff
@@ -9,15 +12,68 @@ class CRM_MembershipExtras_Service_MembershipEndDateCalculator {
    * @param int $membershipId
    *
    * @return string
+   * @throws \CiviCRM_API3_Exception
+   * @throws \CiviCRM_API3_Exception|\Exception
    */
   public static function calculate($membershipId) {
-    $membershipDetails = civicrm_api3('Membership', 'get', [
+    $newEndDate = 'null';
```
erawat, line 19: @MiyaNoctem  'null' not NULL or ''?

CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
```diff
@@ -9,15 +12,68 @@ class CRM_MembershipExtras_Service_MembershipEndDateCalculator {
...
+  /**
+   * Calculates the previous end date for the given membership.
+   *
+   * @param int $membershipID
+   *
+   * @return string
+   * @throws \CiviCRM_API3_Exception|\Exception
+   */
+  public static function calculatePreviousEndDate($membershipID) {
+    $previousEndDate = 'null';
```
erawat, line 40: Same as above

CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
```diff
@@ -9,15 +12,68 @@ class CRM_MembershipExtras_Service_MembershipEndDateCalculator {
...
+    $membershipDetails = self::getMembership($membershipID);
+    $interval = self::getMembershipPeriodInterval($membershipDetails);
+
```
erawat, line 43: I don't think you need a new line here, should be consistent with calculate() method.

CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
```diff
@@ -9,15 +12,68 @@ class CRM_MembershipExtras_Service_MembershipEndDateCalculator {
    * @param int $membershipId
    *
    * @return string
+   * @throws \CiviCRM_API3_Exception
+   * @throws \CiviCRM_API3_Exception|\Exception
    */
   public static function calculate($membershipId) {
```
erawat, line 18: Some if the methods use $membershipId, some use $membershipID . I understand that they can define differently but should it be consistent for the entire class?

erawat (changes requested):

CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
MiyaNoctem: Yeah, has to be `'null'` so CiviCRM sets it as NULL in the DB. `NULL` will not add it to the query, and `''` will set it to an empty string.

CRM/MembershipExtras/Service/MembershipEndDateCalculator.php
MiyaNoctem: Yep, has to be `'null'`. It's a CiviCRM thing.

erawat (approved):

PR #336, opened 2021-01-18
MAE-380: Monthly instalments

erawat:
## Overview

This PR adds: 

- Calculator for fixed period monthly instalments. 
- Calculator for fixed period annual instalments
- Logics to calculate number of instalment based on membership type configuration (for example, 1 instalment for rolling monthly /lifetime  duration for rolling membership type, duration for fixed period membership type when schedule is monthly) 
- Add additional validations to validate given membership types. 
- Add unit tests 
- Add php dockblock for existing functions.


tests/phpunit/CRM/MembershipExtras/Service/MembershipPeriodType/FixedPeriodTypeMonthlyCalculatorTest.php
```diff
@@ -0,0 +1,41 @@
+<?php
+
+use CRM_MembershipExtras_Service_MembershipPeriodType_AbstractBaseTest as BaseTest;
```
MiyaNoctem, line 3: Maybe we should send this class to `CRM_MembershipExtras_Test_Base_Service_...Xxxx`?

I've had to duplicate code in test classes, would be cool to have some place where we can put code we need for more than a single test class. Plus it's not actually a service, its a test.

CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php
```diff
@@ -137,7 +145,16 @@ private function getMembershipStartDate(DateTime $startDate = NULL, DateTime $en
    */
   private function calculateInstalmentAmount() {
     if ($this->membershipTypes[0]->period_type == 'fixed') {
-      $instalmentAmount = new InstalmentAmount(new FixedPeriodCalculator($this->membershipTypes));
+      if ($this->schedule == self::MONTHLY) {
```
MiyaNoctem, line 148: Maybe this could be cleaned up by refactoring to a new method.

Eg.
```php
private function calculateInstalmentAmount() {
  $instalmentAmount = $this->getInstalmentAmountObject();
  $instalmentAmount->getCalculator()->calculate();
  ...
}

private function getInstalmentAmountObject() {
  if ($this->membershipTypes[0]->period_type == 'fixed') {
    if ($this->schedule == self::MONTHLY) {
        $fixedPeriodTypeMonthlyCalculator = new FixedPeriodTypeMonthlyCalculator($this->membershipTypes);
        $fixedPeriodTypeMonthlyCalculator->setStartDate($this->startDate);
        $instalmentAmount = new InstalmentAmount($fixedPeriodTypeMonthlyCalculator);
    }
    else {
      ...
    }
  }
  
  return $instalmentAmount;
}
```

tests/phpunit/CRM/MembershipExtras/Service/MembershipPeriodType/FixedPeriodTypeMonthlyCalculatorTest.php
erawat: That's right, If in this case, we better remove it then. I don't want to send this to CRM_MembershipExtras_Test_Base_Service_* as it will make the test too complex which would make the test difficult to maintain and debug as well as we may introduce the bugs in the tests. 

Perhaps, in the future, we would some out with some nice traits to help us fabricate these test objects.

CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php
erawat: @MiyaNoctem  nice imrovment!

MiyaNoctem (approved):

PR #343, opened 2021-02-10
MAE-352: Pro rated membership with price set

erawat:
## Overview

This PR fixes the issue when we set up a membership with price set and the membership is pro rated, the contribution's line item will still hold the original full amount while the total amount of the contribution is the correct pro rated amount. 

This PR also fixes another issue where tax amount is not pro rated when price set is used with fixed period membership type. 

Below before and after screenshots were created with the same membership types and price sets where

- Membership types are fixed period type. 
- Period start day is 01 Jan and Period roll over day is 31 Dec.
- Duration is 1 year.
- Annual Pro-rata calculation is by months.
- Membership start date is today (10-Feb-2021)

## Before

Pro rated amounts show on membership sign up page

screenshot-compuclient local-2021 02 10-14_26_15

View contribution shows original  line item and  tax amount and after membership is created

screenshot-compuclient local-2021 02 10-14_27_12

## After

Pro rated amounts show on membership sign up page

screenshot-compuclient local-2021 02 10-14_12_53

Pro rated  line item and tax amount show on view contribution after membership is created

screenshot-compuclient local-2021 02 10-17_38_33

## Technical Details

There are few technical points that worth mention are: 

1. Line item amounts will only be pro rated if price set is used and membership type is fixed period membership type.
2. The Logic of altering the line item remains are it was except for the above criteria. 
2. This PR also refactors the Membership payment plan pre hook code to separate classes for contribution and line item object, so each hook class has a single responsibility.   
3.  Uni tests were added to test new implementation of altering line item as well as existing altering line item logic.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/LineItem.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/LineItem.php
@@ -1,6 +1,10 @@
 <?php
 
 use CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_AbstractProcessor as AbstractProcessor;
+use CRM_MembershipExtras_Service_MembershipPeriodType_FixedPeriodTypeAnnualCalculator as FixedPeriodTypeAnnualCalculator;
+use CRM_MembershipExtras_Service_MembershipPeriodType_FixedPeriodTypeMonthlyCalculator as FixedPeriodTypeMonthlyCalculator;
+use CRM_Member_BAO_MembershipType as MembershipType;
+use CRM_MembershipExtras_Service_MembershipInstalmentAmount as InstalmentAmount;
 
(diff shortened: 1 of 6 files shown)


CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/LineItem.php
```diff
@@ -17,12 +21,59 @@ public function __construct(&$params) {
...
-
     if (!empty($this->params['tax_amount'])) {
       $this->params['tax_amount'] = $this->calculateSingleInstalmentAmount($this->params['tax_amount']);
     }
   }
 
+  /**
+   * Calculates pro rata amounts for line item or fixed period membership type
+   * base on payment plan schedule, membreship type, and membership start date.
+   * *
```
ahed-compucorp, line 50: Should change this comment to `Calculates pro rata amounts for line item for fixed period membership type`.

ahed-compucorp (changes requested):

ahed-compucorp (approved):

PR #352, opened 2021-02-22
MAE-479: Alter contribution receive date

erawat:
## Overview

Membership Extras version 5 hides contribution received date field on the user interface by default and the default value also set to sign up date. 

This PR adds logic to alter contribution receive date param to always use membership start date when creating payment plan on civicrm pre hook with contribution object

The contribution receive date param is used as

- A received date for first contribution in a payment plan
- Payment plan start date 

## Before

- First contribution receive date in the payment plan is always set as sign up date

## After

- First contribution receive date in the payment plan will always be set as membership start date 

## Technical Details

Alter receive date param in the pre hook is preferred to create a logic in the UI set receive date as membership start date or membership join date as we can ensure that the first contribution receive date will always set as membership start date as well as the test can be done via unit test.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
@@ -35,8 +35,20 @@ class CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_Contribution 
    * depending on the instalments count.
    */
   public function createPaymentPlan() {
+    $this->alterRecievedDate();
     $this->createRecurringContribution();
     $this->alterFirstContributionParameters();
+  }
+
+  /**
+   * Set Contribution receive date as Membership start date.
+   */
+  private function alterRecievedDate() {
+    $membership = civicrm_api3('Membership', 'getSingle', [
+      'sequential' => 1,
+      'id' => $this->membershipId,
+    ]);
+    $this->params['receive_date'] = $membership['start_date'];
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jamienovick:
Does direct debit start date logic interact with this? @erawat thanks

erawat:
> Does direct debit start date logic interact with this? @erawat thanks

@jamienovick  yes, it interacts with this.  The hook that alters DD received date will trigger and alter this.

jamienovick:
Ok, is there anyway to bake in some tests around that to make sure that the two hooks don't fire in a different order if Civi changes something?

erawat:
> Ok, is there anyway to bake in some tests around that to make sure that the two hooks don't fire in a different order if Civi changes something? @erawat

@jamienovick the hook that alters DD contribution received date or any other extensions is dispatched inside this hook and always runs before creating recurring contribution which after the received date is altered with membership start date.

CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
```diff
@@ -35,10 +35,22 @@ public function getRecurringContribution() {
...
   public function createPaymentPlan() {
+    $this->alterRecievedDate();
     $this->createRecurringContribution();
     $this->alterFirstContributionParameters();
   }
 
+  /**
+   * Set Contribution receive date as Membership start date.
+   */
+  private function alterRecievedDate() {
```
omarabuhussein, line 46: this does not seem that it will work well when we renew with "don't update membership start date on renewal" setting

omarabuhussein (changes requested):

PR #353, opened 2021-02-23
MAE-479:  Payment plan start date

erawat:
## Overview

Membership Extras version 5 hides contribution received date field on the user interface by default and the default value also set to sign up date. 

This PR adds logic in the interface to assign membership start date onto payment plan start date (first contribution received date) automatically when the payment plan is selected. 

Since Membership extras version 5 is using membership start date as a first contribution received date when creating a new membership, the validation to check that membership start date is required is added to the form on the existing validation Form hook. 

## Before

- First contribution receive date in the payment plan is always set as sign up date.
- Membership can be created without select membership start date (civicrm will assign start date automatically based on membership type) 

## After

- First contribution receive date in the payment plan will always be set as membership start date.
- Membership will only be created only when membership start date is selected or the form will throw an error if membership start date is not set. 

## Technical Details

The logic to set receive date as membership start date is done on UI because if  we use pre hook to alter contribution receive date param, we could encounter the issue where the first contribution may not always be membership start date when renew the membership

### DIFF
--- a/CRM/MembershipExtras/Hook/ValidateForm/MembershipPaymentPlan.php
+++ b/CRM/MembershipExtras/Hook/ValidateForm/MembershipPaymentPlan.php
@@ -40,9 +40,27 @@ class CRM_MembershipExtras_Hook_ValidateForm_MembershipPaymentPlan {
 
   /**
    * Validates the payment plan form submission
-   * fields when renewing or creating memberships.
+   * fields when creating a membership.
    */
   public function validate() {
+    $this->validateMembershipStartDate();
+    $this->validatePriceSet();
+  }
+
+  /**
+   * Validates that membership start date is entered.
+   */
+  private function validateMembershipStartDate() {
+    if (empty($this->fields['start_date'])) {
+      $this->errors['start_date'] = ts('Start date is required');
+    }
+  }
+
+  /**
+   * Validates selected fixed membermship types when using price set
+   * have that same period type and same period start day.
+   */
+  private function validatePriceSet() {
     $fixedPeriodStartDays = [];
     $periodTypes = [];
     $priceSetID = $this->fields['price_set_id'];
@@ -73,7 +91,6 @@ class CRM_MembershipExtras_Hook_ValidateForm_MembershipPaymentPlan {
     if (!empty($fixedPeriodStartDays) && count($fixedPeriodStartDays) != 1) {
       $this->errors['price_set_id'] = InvalidMembershipTypeInstalment::SAME_PERIOD_START_DAY;
     }
-
   }
 
   /**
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/ValidateForm/MembershipPaymentPlan.php
```diff
@@ -40,9 +40,20 @@ public function __construct(CRM_Member_Form &$form, &$fields, &$errors) {
...
    * Validates the payment plan form submission
-   * fields when renewing or creating memberships.
+   * fields when creating a membership.
    */
   public function validate() {
+    $this->validateMembershipStartDate();
+    $this->validatePriceSet();
+  }
+
+  private function validateMembershipStartDate() {
```
omarabuhussein, line 50: can you add a dockblock describing what kind of validation it does.

CRM/MembershipExtras/Hook/ValidateForm/MembershipPaymentPlan.php
```diff
@@ -40,9 +40,20 @@ public function __construct(CRM_Member_Form &$form, &$fields, &$errors) {
...
+    if (empty($this->fields['start_date'])) {
+      $this->errors['start_date'] = ts('Start date is required');
+    }
+  }
+
+  private function validatePriceSet() {
```
omarabuhussein, line 56: same here

membershipextras.php
```diff
@@ -369,9 +369,8 @@ function membershipextras_civicrm_pageRun($page) {
 function membershipextras_civicrm_validateForm($formName, &$fields, &$files, &$form, &$errors) {
   $formAction = $form->getAction();
   $isNewMembershipForm = ($formName === 'CRM_Member_Form_Membership' && ($formAction & CRM_Core_Action::ADD));
-  $isRenewMembershipForm = ($formName === 'CRM_Member_Form_MembershipRenewal' && ($formAction & CRM_Core_Action::RENEW));
   $isPaymentPlanPaymentlanPayment = CRM_MembershipExtras_Utils_InstalmentSchedule::isPaymentPlanWithSchedule();
-  if (($isNewMembershipForm || $isRenewMembershipForm) && $isPaymentPlanPaymentlanPayment) {
+  if ($isNewMembershipForm && $isPaymentPlanPaymentlanPayment) {
```
omarabuhussein, line 373: can we change this variable $isPaymentPlanPaymentlanPayment name to $isPaymentPlanPayment

templates/CRM/Member/Form/PaymentPlanToggler.tpl
```diff
@@ -91,16 +91,27 @@
...
         if ($('#payment_plan_schedule_row').is(":hidden")) {
           return;
         }
         generateInstalmentSchedule(isPriceSetSelected());
+        assignReceiveDate();
       });
     }
 
+    /**
+     * Assigns payment plan received date from either start date or join date
```
omarabuhussein, line 107: this assigns first contribution receive date and not the payment plan, can you also change the function name to assignFirstContributionReceiveDate

omarabuhussein (changes requested):

omarabuhussein (approved):

PR #356, opened 2021-02-26
IO-121: Save settings when creating new membership type

erawat:
## Overview

This PR fixes the issue when extended membership type settings have not been saved when creating a new membership type. 

The PR also adds script to the template to hide fixed rollover days when selected fixed membership type to ensure that the field will not show when the page returns from validation if the user selects month as a duration interval for fixed period membership type. 

## Before

- Membership type colour and annual pro rata calculation values are only created  into the settings on edit mode. 

## After

- Membership type colour and annual pro rata calculation values are created and updated on when create or update membership type. 

## Technical Details

A form object that passes on  CiviCRM postProcess hook does not contain an ID when creating a new object, so the membership  type settings will never be saved. 

In order to address this issue, we check if the ID does not exist, then we will call Membership Type API to get membership type ID by name as membership type name is unique. By doing this, we can ensure that the settings will be saved when creating new membership type.

### DIFF
--- a/CRM/MembershipExtras/Hook/PostProcess/MembershipTypeSetting.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/MembershipTypeSetting.php
@@ -48,13 +48,12 @@ class CRM_MembershipExtras_Hook_PostProcess_MembershipTypeSetting {
       $this->settings = [];
     }
 
-    if (!empty($this->form->_id)) {
-      $annualProRataCalculationValue = $this->form->_submitValues['membership_type_annual_pro_rata_calculation'];
-      $this->settings[$this->form->_id]  = [
-        'membership_type_annual_pro_rata_calculation' => $annualProRataCalculationValue,
-      ];
-      Civi::settings()->set(SettingsManager::MEMBERSHIP_TYPE_SETTINGS_KEY, $this->settings);
-    }
+    $membershipTypeId = $this->getMembershipTypeId();
+    $annualProRataCalculationValue = $this->form->_submitValues['membership_type_annual_pro_rata_calculation'];
+    $this->settings[$membershipTypeId]  = [
+      'membership_type_annual_pro_rata_calculation' => $annualProRataCalculationValue,
+    ];
+    Civi::settings()->set(SettingsManager::MEMBERSHIP_TYPE_SETTINGS_KEY, $this->settings);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Hook/PostProcess/MembershipTypeSetting.php
```diff
@@ -48,13 +48,13 @@ private function updateMembershipTypeSettings() {
...
     }
 
-    if (!empty($this->form->_id)) {
-      $annualProRataCalculationValue = $this->form->_submitValues['membership_type_annual_pro_rata_calculation'];
-      $this->settings[$this->form->_id]  = [
-        'membership_type_annual_pro_rata_calculation' => $annualProRataCalculationValue,
-      ];
-      Civi::settings()->set(SettingsManager::MEMBERSHIP_TYPE_SETTINGS_KEY, $this->settings);
-    }
+    $membershipTypeId = !empty($this->form->_id) ? $this->form->_id : $this->getMembershipTypeId();
```
ahed-compucorp, line 51: Should we move the condition to `getMembershipTypeId` method.

templates/CRM/Member/Form/MembershipType/Settings.tpl
```diff
@@ -49,6 +49,7 @@
      * Handles fields when Membership Type fixed period is selected
     */
     function handleFixedPeriod() {
+      $('#month_fixed_rollover_day_row').hide();
```
ahed-compucorp, line 52: Could you mention this in the PR description too?

CRM/MembershipExtras/Hook/PostProcess/UpdateMembershipTypeColour.php
```diff
@@ -48,16 +48,16 @@ private function updateMembershipTypeColourFields() {
...
     }
 
-    if (!empty($this->form->_id)) {
-      $setMembershipColour = $this->form->_submitValues['set_membership_colour'];
-      $membershipColour = $this->form->_submitValues['membership_colour'];
-      $membershipTypeColourSettings[$this->form->_id] = [
-        'set_membership_colour' => $setMembershipColour,
-        'membership_colour' => $setMembershipColour ? $membershipColour : ''
-      ];
+    $membershipTypeId = !empty($this->form->_id) ? $this->form->_id : $this->getMembershipTypeId();
```
ahed-compucorp, line 51: The same note, Should we move the condition to `getMembershipTypeId` method?

ahed-compucorp (changes requested):

CRM/MembershipExtras/Hook/PostProcess/MembershipTypeSetting.php
erawat: @ahed-compucorp  I was thinking about this too, but I do not want to call the API when we edit the membership form as the ID would already exist.

ahed-compucorp: I mean like this :


    private function getMembershipTypeId() {
        if (!empty($this->form->_id)) {
           $membershipTypeId = $this->form->_id; 
        } else { 
          $membershipTypeId = civicrm_api3('MembershipType', 'getsingle', [
            'return' => ['id'],
            'name' => $this->form->exportValues()['name'],
          ])['id'];
        }
        return $membershipTypeId;
    }

ahed-compucorp (approved):

PR #358, opened 2021-03-04
IO-139: Add Instalment line item(s) to schedule 

erawat:
## Overview

Total amount and tax amount on currently screen is confusing.  This PR adds line item(s) as sub table on each instalment to visualize financial type, quantity, unit price, sub total, tax amount as total amount similar to CiviCRM contribution line item. 

Line items table works with 

- Rolling membership type
- Fixed period membership type
- Price set with membership type
- Price set with none membership type 

## Before

Peek 2021-03-04 10-27

## After

Peek 2021-03-04 10-16

## Technical Details

Previously, the instalments display was generated each time by rewriting DOM, this method was expensive especially when we need to rewrite the DOM with nested table each time and the display is not reusable.  

Since, we need a complex table for displaying instalments, the CiviCRM page for displaying the content is created and the membership form is loading the instalments table by using [CRM.loadPage()](https://docs.civicrm.org/dev/en/latest/framework/ajax/#crmloadpageurl-options) function that provided by CiviCRM to inject the page into the membership form. This method is widely used in CiviCRM in many pages.  

Finally, we have also changed the way to display instalments schedule, the API format has also been updated to remove formatting like currency symbol, status label, so the API will return raw data and the page will handle the display format. This way makes the API result more flexible to reuse in other systems or a module like webform.

### DIFF
--- a/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php
+++ b/CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php
@@ -1,6 +1,5 @@
 <?php
 
-use CRM_MembershipExtras_Service_MoneyUtilities as MoneyUtilities;
 use CRM_MembershipExtras_Exception_InvalidMembershipTypeInstalment as InvalidMembershipTypeInstalment;
 use CRM_MembershipExtras_DTO_ScheduleInstalmentAmount as ScheduleInstalmentAmount;
 use CRM_MembershipExtras_Service_MembershipInstalmentAmount as InstalmentAmount;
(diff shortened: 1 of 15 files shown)


jamienovick:
> Since we have changed the way to display instalments schedule, the API format has also been updated to remove formatting like currency symbol, status label, so the API will return raw data and the page will handle the display format. This way makes the API result more flexible to reuse in other systems or a module like webform.

Can we just confirm that this works with sites that use Euros or other currencies i.e. for FEMS or WHF?

erawat:
> > Since we have changed the way to display instalments schedule, the API format has also been updated to remove formatting like currency symbol, status label, so the API will return raw data and the page will handle the display format. This way makes the API result more flexible to reuse in other systems or a module like webform.
> 
> Can we just confirm that this works with sites that use Euros or other currencies i.e. for FEMS or WHF?

@jamienovick  This will work with any currency as the template is using default currency symbol and date format from civicrm localisation settings.

CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php
```diff
@@ -265,10 +325,23 @@ private function getInstalmentsTotalAmount(array $instalments) {
...
 
+  /**
+   * Calculates Single Instalment Amount
+   *
+   * @param $amount
+   * @param $divisor
+   * @return float|int
+   */
+  private function calculateSingleInstalmentAmount($amount, $divisor) {
+    return $amount / $divisor;
```
ahed-compucorp, line 344: Should we call `round` here with `precision = 2` by default ?

tests/phpunit/api/v3/PaymentSchedule/GetPaymentScheduleTest.php
```diff
@@ -106,8 +97,8 @@ public function testGetMonthlyFixedMembershipTypeCalculatedByDays() {
     //No of months between start date 01 Jan to membership rollover day is 9 months
     //2021 has 365 days
     $expectedAmount = round(((120 / 365) * $durationInDays) / 9, 2);
-    $expectedAmount = $this->currencySymbol . $expectedAmount;
-    $expectedTaxAmount = $this->currencySymbol . 0;
+    $expectedAmount = $expectedAmount;
```
ahed-compucorp, line 100: Should we remove this line?

tests/phpunit/api/v3/PaymentSchedule/GetPaymentScheduleTest.php
```diff
@@ -128,8 +119,8 @@ public function testGetAnnualFixedMembershipTypeCalculateByMonths() {
     $expectedAmount = (120 / 12) * 9;
     $instalments = $this->getMembershipTypeSchedule($membershipType['id'], 'annual', $formattedStartDate);
     $this->assertCount(1, $instalments['instalments']);
-    $expectedAmount = $this->currencySymbol . $expectedAmount;
-    $expectedTaxAmount = $this->currencySymbol . 0;
+    $expectedAmount = $expectedAmount;
```
ahed-compucorp, line 122: Should we remove this line?

ahed-compucorp (changes requested):

CRM/MembershipExtras/Service/MembershipInstalmentsSchedule.php
erawat: @ahed-compucorp  I did not round up there because I want to leave the amount as it is so we will only round up when we need it as it see that the round up function getting duplicated everywhere.

ahed-compucorp (approved):

PR #360, opened 2021-03-08
MAE-484: Calculation for fixed membership with past year start date

erawat:
## Overview

This PR adds functionality to support pro-rata calculation for fixed period membership type where membership start date is in the part. 

For example, 

Membership type configuration

- Membership start day: 01 Dec
- Membership rollover days: 30 Nov
- Membership Duration 1 Year

Membership start date: 01/01/2020
Membership end date: user not select

### Before

As above example, 

The membership end date will default to 30/Nov/2021 and the different in days will be 700 days so the pro-rata will be calculated based on 700 days  if annual configuration is by days

### After
 
The membership end date will default to 30/Nov/2001 and the different in days will be 335 days so the pro-rata will be calculated based on 335 days if annual configuration is by days


CRM/MembershipExtras/Service/MembershipPeriodType/FixedPeriodTypeCalculator.php
```diff
@@ -57,15 +65,23 @@ public function calculate() {
...
+          $this->recalCalcuateEndDate();
+        }
+        $diff = $membershipTypeDurationCalculator->calculateMonthsBasedOnDates($this->startDate, $this->endDate, $this->joinDate);
       }
       else {
         $duration  = $membershipTypeDurationCalculator->calculateOriginalInDays();
-        $diff = $membershipTypeDurationCalculator->calculateDaysBasedOnDates($this->startDate);
+        $diff = $membershipTypeDurationCalculator->calculateDaysBasedOnDates($this->startDate, $this->endDate, $this->joinDate);
+        if ($this->isDurationWithInOneYearPeriod($duration, $diff)) {
+          $this->recalCalcuateEndDate();
```
ahed-compucorp, line 81: If the end date was more by 3 years. Should we handle that case in `recalCalcuateEndDate`?

erawat: CiviCRM core actually calculates end date for you only 2 years, so your start date is 2018, you will end date will be 2020 so 2020 - 1 is 2019, which should cover this scenario. The scenario is a unique for testing purpose renewal only.

ahed-compucorp (approved):

ahed-compucorp (approved):

PR #363, opened 2021-03-09
MAE-474: Auto filling membership start date

erawat:
## Overview

Currently, CiviCRM does not set membership start date automatically when the form is loaded.  When the payment plan is selected without membership start date entered, today date will be used by default, but membership start date is still left empty, and it can be confused. This PR fixes that. 

This PR also changes the API request on the payment plan template to exclude the end date from the request because the end date shall be calculated automatically based on membership type and to avoid, user to input incorrect membership end date.

## Before

- Start date is not set when the payment schedule is generated.
- End date can be passed to API.

## After

- Start date is always set when the payment schedule is generated.
- End date is set automatically from the API.


templates/CRM/Member/Form/PaymentPlanToggler.tpl
```diff
@@ -215,6 +214,8 @@
      * Sets membership dates
      */
     function setMembershipDates(startDate, endDate) {
+      console.log(startDate);
+      console.log(endDate);
```
ahed-compucorp, line 218: Should we remove those log lines ?

PR #368, opened 2021-03-12
MAE-492: Fix membership end date after CiviCRM 5.35 update

ahed-compucorp:
## Overview
After updating to CiviCRM 5.35, If we created a payment plan membership with pending status, we found out that the end date will increase each time we recorded a new payment.

## Before
Recording a new payment **will increase** the membership end date.
Peek 2021-03-12 21-12

## After
Recording a new payment **will not increase** the membership end date.
Peek 2021-03-12 20-47

## Technical Details
This PR https://github.com/civicrm/civicrm-core/pull/18410  to CiviCRM core has the [following change](https://github.com/civicrm/civicrm-core/pull/18410/commits/cbd83ddebb3c7e88157f90905c4ac63bf5214c44) : 

```diff
- $prevnext[] = $this->createElement('submit', $buttonName, $button['name'], $attrs);
+ $attrs['type'] = 'submit';
+ $prevnext[] = $this->createElement('xbutton', $buttonName, $button['name'], $attrs);
```

Instead of creating a button with the value "Record Payment" like this
```html
<input class="crm-form-submit default validate" onclick="return verify( );" crm-icon="fa-check" name="_qf_AdditionalPayment_upload" value="Record Payment" type="submit" id="_qf_AdditionalPayment_upload">
```

It will create a button with the value "1" like this
```html
<button class="crm-form-submit default validate crm-button crm-button-type-upload crm-button_qf_AdditionalPayment_upload" onclick="return verify( );" value="1" type="submit" name="_qf_AdditionalPayment_upload" id="_qf_AdditionalPayment_upload"><i aria-hidden="true" class="crm-i fa-check"></i> Record Payment</button>
```

And the condition [here](https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/7acc3dadd384d03e1d7c4d32313a0715db64b776/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php#L89) will be false
```php
$isRecordPayment = CRM_Utils_Request::retrieve('_qf_AdditionalPayment_upload', 'String') === 'Record Payment';
```

The main point is to figure out if the submitted form is a record payment submission or refund submission and we can do that by checking the the variable `_paymentType` in the `AdditionalPayment` Form class and prepare it for `MembershipEdit` class.

## Comments
Another potential solution is to alter the button attributes and inject the value "Record Payment" again like this

```php
/**
 * Implements hook_civicrm_buildForm().
 */
function membershipextras_civicrm_buildForm($formName, &$form) {
  if ($formName == 'CRM_Contribute_Form_AdditionalPayment') {
    $buttons = $form->getElement('buttons');
    $elements = $buttons->getElements();
    $elements[0]->setValue("Record Payment");
  }
}
```

### DIFF
--- a/membershipextras.php
+++ b/membershipextras.php
@@ -177,7 +177,8 @@ function membershipextras_civicrm_pre($op, $objectName, $id, &$params) {
   }
 
   if ($objectName === 'Membership' && $op == 'edit') {
-    $membershipPreHook = new CRM_MembershipExtras_Hook_Pre_MembershipEdit($id, $params, $contributionID);
+    $paymentType = Civi::$statics[E::LONG_NAME]['paymentType'] ?? '';
+    $membershipPreHook = new CRM_MembershipExtras_Hook_Pre_MembershipEdit($id, $params, $contributionID, $paymentType);
     $membershipPreHook->preProcess();
   }
 
@@ -333,6 +334,10 @@ function membershipextras_civicrm_buildForm($formName, &$form) {
   if ($formName === 'CRM_Contribute_Form_Contribution') {
     $membershipTypeHook = new CRM_MembershipExtras_Hook_BuildForm_ContributionEdit();
     $membershipTypeHook->buildForm();
+  }
+
+  if ($formName == 'CRM_Contribute_Form_AdditionalPayment') {
+    Civi::$statics[E::LONG_NAME]['paymentType'] = $form->getVar('_paymentType');
   }
 }
 
--- a/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
@@ -30,6 +30,15 @@ class CRM_MembershipExtras_Hook_Pre_MembershipEdit {
   private $paymentContributionID;
 
   /**
+   * The payment type e.g. refund or owed
+   *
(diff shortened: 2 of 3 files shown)


omarabuhussein (approved):

membershipextras.php
```diff
@@ -177,7 +177,8 @@ function membershipextras_civicrm_pre($op, $objectName, $id, &$params) {
   }
 
   if ($objectName === 'Membership' && $op == 'edit') {
-    $membershipPreHook = new CRM_MembershipExtras_Hook_Pre_MembershipEdit($id, $params, $contributionID);
+    $paymentType = Civi::$statics[E::LONG_NAME]['paymentType'] ?? '';
+    $membershipPreHook = new CRM_MembershipExtras_Hook_Pre_MembershipEdit($id, $params, $contributionID, $paymentType);
     $membershipPreHook->preProcess();
   }
```
ahed-compucorp, line 183: Just removed this line at line 182
```diff
- unset(Civi::$statics[E::LONG_NAME]['paymentType']);
```
because if you created a pending memberships in one go from a priceset then this `if` condition will be true for each membership and we need to keep this `Civi::$statics[E::LONG_NAME]['paymentType']` for the other memberships

omarabuhussein (approved):

PR #366, opened 2021-03-11
MAE-475: Display instalment summary amount and prorated number

erawat:
## Overview

This PR adds the schedule summary amount to the instalments table.

This PR also improves the code and templates by:

- Using CiviCRM tax term for displaying tax text instead of hard code in to the template.
- Separating JavaScript, stylesheet from the template including existing payment toggle. 
- Adding additional tests as well as add Instalment page test.  
- Refactor the class name, so it has a meaningful name. 

## Before

Subtotal, tax amount,  total amount and prorated unit (if fixed membership type is selected) were not shown on the table. 

## After

Subtotal, tax amount,  total amount and prorated unit (if fixed membership type is selected) are displayed on the table. 

Example of what the summary amount look like on different membership types. 

**Prorated by months**
screenshot-compuclient local-2021 03 11-11_36_56

**Prorated by days**
screenshot-compuclient local-2021 03 11-11_36_25

**Rolling membership** 
screenshot-compuclient local-2021 03 11-11_36_01

PR #374, opened 2021-03-30
MAE-517: Fix schedule instalment display on non payment plan tab

erawat:
## Overview

This PR fixes the issue when select membership type then instalments schedule table is shown on the non payment plan tab (contribution). 

## Before

Instalment Schedule displays on the contribution tab.

01a70289-0574-4be4-bc50-3949e66d9f84

## After

Instalment Schedule displays ONLY when select payment plan tab.

Peek 2021-03-30 11-38

## Technical Details

Since we refactored the code and separate the javascript to the different file in this [PR](https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/366). 

The constant variables that requires to exist when smart template is completed have not been passed to the javascript in the separate file because they are the indifferent scope therefore, the screen are not working correctly

_How this was addressed?_ 

In order to maintain the javascript logic from the smart template and javascript logic relies on the constant values on the template, we would need to put the function in jquery anonymous function inside the global scope function so it can be called from anywhere. 

Alternatively, we could define a namespace to reduce risk of name collisions instead of using a global scope function but in this case, we are less likely to run into any issue as this function will only be used for payment toggler screen.

### DIFF
--- a/js/paymentPlanToggler.js
+++ b/js/paymentPlanToggler.js
@@ -1,298 +1,304 @@
-CRM.$(function ($) {
-  /**
-   * Perform changes on form to add payment plan as an option to pay for
-   * membership.
-   */
-  $(function () {
-    moveMembershipFormFields();
-    setMembershipFormEvents();
-    initializeMembershipForm();
-  });
+function paymentPlanToggler(togglerValue, currencySymbol) {
 
-  /**
-   * Leaves form in a consistent state after the fields have been moved and the
-   * new events added.
-   */
-  function initializeMembershipForm () {
-    selectPaymentPlanTab(togglerValue);
-  }
+  CRM.$(function ($) {
 
-  /**
-   * Reorders fields within the form to follow a coherent logic with the new
-   * payment plan fields.
-   */
-  function moveMembershipFormFields () {
-    $('#contributionTypeToggle').insertBefore($('#is_different_contribution_contact').parent().parent());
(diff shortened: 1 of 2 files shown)


js/paymentPlanToggler.js
```diff
@@ -1,298 +1,304 @@
...
+function paymentPlanToggler(togglerValue, currencySymbol) {
 
-  /**
-   * Leaves form in a consistent state after the fields have been moved and the
-   * new events added.
-   */
-  function initializeMembershipForm () {
-    selectPaymentPlanTab(togglerValue);
-  }
+  CRM.$(function ($) {
```
ahed-compucorp, line 3: Should we remove this line ?

ahed-compucorp (changes requested):

js/paymentPlanToggler.js
erawat: We would like to keep it as it is to ensure that elements are loaded and they are in CRM. context.

ahed-compucorp: You use it inside `CRM.$(function ($) {` so you have the assurance , It should be fine.

erawat: @ahed-compucorp  It needs a closure here otherwise, I would need to replace $ with CRM.$

ahed-compucorp (approved):

PR #376, opened 2021-04-07
MAE-523: Improve/fix instalment amount calculations.

erawat:
## Overview

This PR does the followings: 

- Fixed CiviCRM 5.35.1 compatibility. 
- Fixed precision issue on the calculation. 

Issue 1: CiviCRM 5.35.1 

In CiviCRM 5.35,  `membership_id` will not add to parameter when creating contribution / line item objects [Membership.php#L1451)](https://github.com/civicrm/civicrm-core/blob/5.35.0/CRM/Member/Form/Membership.php#L1451), and `membership_payment` record will only be created once the contribution is saved.  However, Membership extras require membership_id to calculate instalment amount and frequency.   

Issue 2: Precision

Membership extras version 5, provides a new UI to visualise the instalment details before the membership is created.  

screenshot-compuclient local-2021 04 14-15_17_57

The instalment details are included instalment amount, tax, total amount as well as sub line item that show the line item(s) of the instalment.

Since the precision error can happen easily when we pro-rated amount by months or by days, see  one of the error that could happen as  example below when membership is created with a price set. 

```
Price set 1
Membership 1 fee  = £10.55 / annual
Membership 2 fee  = £10 / annual
Membership fee total = £22.55 / annual

if we prorated for 269 days and pay 9 instalments the membership fee would be  £18.18 / annual
Sub total amount = 15.15
Tax amount = 3.03
Total Amount = 18.18

Instalment amount is 18.18 / 9 = 2.02

Line items would look like this 
Line Item 1 amount = 7.78
Line Item 1 tax = 1.56
Line Item 1 amount + tax = 9.34

Line Item 2 amount = 7.37
Line Item 2 tax = 1.47
Line Item 2 amount + tax = 8.84

Line1 + Line 2 = 18.18
Line item total amount  is 18.18 / 9 = 2.02

However if we calculate by divide line item (or instalment amount) by number of instalment, we would have different result as below. 

Line Item 1 per instalment  would be 9.34 / 9 = 1.03
Line Item 1 per instalment would be  8.84 / 9 = 0.98

Instalment amount would be  1.03 + 0.98 = 2.01
```

The different would be £0.01 when we created an instalment because CiviCRM records amounts in different places for instance, contribution, contribution line item, membership line item, financial line item, recurring contribution. 

The calculation that is currently being  implemented in the PR is that total amount is calculated from line item, we do it this way because CiviCRM records financial line item this way, and then we would need to ensure that the calculation will be done the same when we virtualise the schedule and when contribution and line items are created, so this PR does that. 

## Before

- Calculation was incorrectly because `membership_id` did not exist in the contribution / line item object since we upgrade to CiviCRM 5.35.1
- The calculation was done separately between when we virtualise the schedule and when we create a contribution / line item. 

## Technical Details

There are a couple of things that worth mention in this section. 

- In this PR, I refactored the code that used to calculate the instalment amount to MembershipInstalmentAmountCalculator class, so we can reuse the calculation methods in the hook and the instalment calculation. 
- Since Membership Payment record does not exist (yet) and membership ID did not pass in the contribution and line item object, we would need to obtain the `membership_id` from the membership line item that attached to the contribution parameter which is done [here ](https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/25c378af028dcc6a42796f34610b6b74ece46e48/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php#L59)
- When creating a line item, we also need to obtain the `membership_id` in order to calculate the right amount for each instalment however, the `membership_id` will not always be available because the line item could be a contribution line item which does not a have `membership_id` attached to it, the solution to address this is to use static class available that it was set when contribution object was being created.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
@@ -3,6 +3,9 @@
 use CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_AbstractProcessor as AbstractProcessor;
 use CRM_MembershipExtras_Hook_CustomDispatch_CalculateContributionReceiveDate as CalculateContributionReceiveDateDispatcher;
 
+/**
+ * Class CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_Contribution
+ */
 class CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_Contribution extends AbstractProcessor {
 
   /**
@@ -12,10 +15,60 @@ class CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_Contribution 
    */
   private $recurringContribution;
 
+  /**
+   * @var array
+   */
+  private $membershipTypeLineItems = [];
+
+  /**
+   * @var array
+   */
+  private $nonMembershipTypeLineItems = [];
+
+  /**
+   * @var CRM_MembershipExtras_DTO_ScheduleInstalmentAmount
+   */
+  private $instalmentAmount;
+
+  /**
+   * @var string
+   */
+  private $periodType;
+
+  /**
+   * @var void
+   */
+  private $instalmentAmountCalculator;
+
+  /**
+   * @throws CiviCRM_API3_Exception
+   * @throws CRM_Core_Exception
+   */
   public function __construct(&$params) {
     $this->params = &$params;
+    $this->setMembershipId();
     $this->assignInstalmentDetails();
-    $this->fixedPeriodMembershipTypes = $this->getLineItemsFixedPeriodMembershipType();
+    $this->handleContributionLineItems();
+    $this->instalmentAmountCalculator = $this->getInstalmentAmountCalculator($this->membershipTypeLineItems, $this->periodType);
+  }
+
+  /**
+   * Sets membership ID from line item.
+   *
+   * Since CiviCRM 5.35.1, membership ID will not be passed
+   * to contribution param, as we would need the membership dates from
+   * membership object to calculate instalment details, in order to
+   * do this, we need to get membership ID from line item entity.
+   */
+  public function setMembershipId() {
+    $lineItems = $this->params['line_item'];
+    array_walk($lineItems, function (&$lineItem, $key) {
+      array_walk($lineItem, function (&$line, $key) {
+        if ($line['entity_table'] == 'civicrm_membership') {
+          self::$membership_id = $line['entity_id'];
+        }
+      });
+    });
   }
 
   /**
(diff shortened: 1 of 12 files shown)


CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
```diff
@@ -116,48 +210,41 @@ private function alterFirstContributionParameters() {
...
-   * Gets only fixed period membeship type that belong to each line item
+   * Assigns membership types from line items.
+   * Assigns non membership price field values from line item.
+   * Sets isFixedMembershipType property
+   *
+   * @throws CRM_Core_Exception
    */
-  private function getLineItemsFixedPeriodMembershipType() {
-    $membershipTypes = [];
+  private function setMembershipTypes() {
```
ahed-compucorp, line 223: Code becomes complicated.
Should we refactor and create another 3 methods ? e.g.
- `setMembershipTypes()` should only set membershipTypes.
- `nonMembershipTypeLineItems(()`
- `adjustMinimumFee()` should run after the `setMembershipTypes()` in the `__construct`
- `setIsFixedMembershipType()`
Or maybe rename the method to a better name to reflect what is going on.

tests/phpunit/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/LineItemTest.php
```diff
@@ -23,8 +23,17 @@ class CRM_MembershipExtras_Hook_Pre_MembershipPaymentPlanProcessor_LineItemTest
...
   private $priceSet;
+  /**
+   * @var mixed
+   */
+  private $membership;
+
+  public function setUp() {
+    //We mocked membership ID as the membership ID is always when
+    //Contribution pre_process hook is called.
+    MembershipPaymentPlanProcessor::$membership_id = $this->membership['id'];;
```
ahed-compucorp, line 35: Should we reset it after the test using the `teardown` method ?

ahed-compucorp (changes requested):

tests/phpunit/CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/LineItemTest.php
erawat: @ahed-compucorp  perhaps, we don't need it as we always reset it when we setup, it is not like we need to tear down environment variables.

CRM/MembershipExtras/Hook/Pre/MembershipPaymentPlanProcessor/Contribution.php
erawat: @ahed-compucorp I have refactored the function by separating each responsibility to each function. I did not consider running method after method on the contractor as we would need to iterate line items each time.

ahed-compucorp (approved):

PR #383, opened 2021-06-02
MAE-522: Prevent extending membership end date

ahed-compucorp:
## Overview
Usually recording the payment for the membership's contribution means :
- Activating the membership by changing its status from pending to active.
- Renewing the membership by extending its end_date.

We have a use case when the admin adding a new line items to a completed contribution and that will change its status to "Partially paid", but recording the payment for the added line items will trigger the renewal membership processing if the contribution was related to a membership.

## Before
Recording a payment for a partially paid contribution will renew its related membership.

Peek 2021-06-02 11-33

## After
Recording a payment for a partially paid contribution will not renew its related membership.

Peek 2021-06-02 11-31

## Technical Details
Usually recording the payment for the membership's contribution means :
- Activating the membership by changing its status from pending to
active.
- Renewing the membership by extending its end_date.

The method `updateMembershipBasedOnCompletionOfContribution` at [Contribution.php#L4793](https://github.com/civicrm/civicrm-core/blob/5.35.2/CRM/Contribute/BAO/Contribution.php#L4793) does not  handle the case when recording another payment for the membership's contribution.

At Compucorp, we use the "biz.jmaconsulting.lineitemedit" extension to
add and edit the line items.

Adding a line item changes the contribution status to "Partially Paid"
and recording a payment for an active membership's contribution will
renew it by default and extend its end_date, but this is undesirable
behaviour and this PR will prevent it.

At first, I tried to check the `MembershipPayment` but that wasn't right because the CiviCRM will create it during  the membership creation or renewing. Then I checked if the membership was active and not `is_pay_later` with checking if the contribution is `Partially paid`. lastly, I tried to check the contribution balance and the paid amount that should be larger than the membership fee , but I couldn't thing of a use case that will need it so I removed it from this PR.

## Comments
It was a bit tricky to decide where to send this PR, let me see what you think about it.

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
@@ -62,6 +62,7 @@ class CRM_MembershipExtras_Hook_Pre_MembershipEdit {
    */
   public function preProcess() {
     if ($this->paymentContributionID || $this->isRecordingPayment() || $this->isBulkStatusUpdate()) {
+      $this->preventEditingTheActiveMembershipIfItsContributionHasAnotherPayment();
       $this->preventExtendingPaymentPlanMembership();
     }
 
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
```diff
@@ -157,6 +158,50 @@ public function preventExtendingPaymentPlanMembership() {
     }
   }
 
+  /**
+   * Prevents editing the active Membership if its contribution has another
+   * payment.
+   *
+   * Usually recording the payment for the membership's contribution means :
```
omarabuhussein, line 165: This is the case only for non payment-plans payments  , it worth noting that here.

CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
```diff
@@ -157,6 +158,50 @@ public function preventExtendingPaymentPlanMembership() {
...
+   * and edit the line items.
+   *
+   * Adding a line item changes the contribution status to "Partially Paid"
+   * and recording a payment for an active membership's contribution will renew
+   * it by default and extend its end_date, but this is undesirable behaviour
+   * and this method will prevent it.
+   */
+  public function preventEditingTheActiveMembershipIfItsContributionHasAnotherPayment() {
+    // Check if the membership is active and is not pay later.
+    $membershipCount = civicrm_api3('Membership', 'getcount', [
```
omarabuhussein, line 179: I am not sure, but I think using "get" is more efficient.

CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
```diff
@@ -157,6 +158,50 @@ public function preventExtendingPaymentPlanMembership() {
...
+      'sequential' => 1,
+      'active_only' => 1,
+      'id' => $this->id,
+      'is_pay_later' => 0,
```
omarabuhussein, line 183: Civi  does not keep this field updated, so if you create a membership with :

1- Pending contribution (is_pay_later will be 1)
2- Complete the contribution (is_pay_later will remain 1)

so if someone then added new line item to the completed contribution then it won't be picked here, since is_pay_later is still = 1 since civi do not update it.

but to be honest, I don't it is needed, so If the membership is : 

1- active
2- and the contribution is currently partially paid and we are trying to record another payment.

then just skip updating its end date, since CiviCRM only extends the date if the payment is completed. But this might cause issues if we are using the line item editor with Payment plans, so I just asked Sheela on the ticket if we are using it with payment plans.

omarabuhussein (changes requested):

CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
ahed-compucorp: Added a condition to skip payment plans.

omarabuhussein: actually, I am rethinking this, and even the current approach might be an issue, here is why:

1- Create a Membership with a completed contribution (the membership status will be active).
2- Renew the membership manually but with a Pending contribution (the membership will remain active, and the end date will not get extended yet since the contribution is pending).
3- Partially pay the contribution (the membership is still active, and the end date still the same since Civi only extends the end date if the contribution got completed).
4- Pay the remaining amount of the contribution, now this step on CiviCRM core will complete the contribution, then extend the membership end date, but with this new logic, it will see an active membership with a partially paid contribution getting completed, and thus it will prevent the end date from getting extended, which is wrong.

at this point I am not sure how can we make this work, will need more thinking.

PR #389, opened 2021-08-20
MAE-581: Prevent updating 'Next contribution date' from outside Membershipextras for offline payment processors

omarabuhussein:
## Overview

I here prevent any update to the recurring contribution 'next_sched_contribution_date' field from outside
Membershipextras (Basically preventing CiviCRM core and the webform_civicrm module from altering its value) since it will be used for autorenewal, and thus we need to be able to control its value with external interference. Hence that this only affect offline payment processors (any payment processor that extends the Payment_Manual class).

## Before

Creating payment plan from the admin form or from a webform will set the `next_sched_contribution_date` value 

bee

## After

Creating payment plan from admin form or from webform will set the `next_sched_contribution_date` to blank

affff

Hence that this is temporary, and I will later implement new code to set its value but form within Membershipextras, since the value that we see in the "Before" section gif is coming from CiviCRM core which we don't control. 

## Technical details

I unset the value by removing the value of this field inside the Recurring Contribution Pre hook, this means that
any API or DAO that try to change the value of this field will not work (which is how CiviCRM core and webform_civicrm module alter its value), Which allow us to control its value within Membershipextras using Direct SQL queries.

### DIFF
--- a/CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
+++ b/CRM/MembershipExtras/Hook/PostProcess/UpdateSubscription.php
@@ -76,11 +76,6 @@ class CRM_MembershipExtras_Hook_PostProcess_UpdateSubscription {
       'auto_renew' => $autoRenew,
     ];
 
-    $nextScheduledDate = CRM_Utils_Array::value('next_sched_contribution_date', $this->formValues);
-    if (!empty($nextScheduledDate)) {
-      $params['next_sched_contribution_date'] = $this->formValues['next_sched_contribution_date'];
-    }
-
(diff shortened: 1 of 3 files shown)


ahed-compucorp (approved):

jamienovick:
Hey, not to complicate things but just wondering how this might interact with other payment processors such as if we were to use Stripe for recurring credit card payments? @omarabuhussein 

Could/do we limit the scope only to our payment processor perhaps?

omarabuhussein:
@jamienovick  this is already the case in this PR, I also have tests to ensure that.

omarabuhussein:
I just updated the title and the PR description to make that more clear.

PR #398, opened 2021-09-28
MAE-612: Changes to autorenewal job trigger and migrating upgrader

omarabuhussein:
## Overview

There was a misunderstanding that the trigger for autorenewal should be the next scheduled contribution date. There is a risk that this date may be before or after the end date of the membership (specially in Direct debit payment plans).

As such we changed the "trigger" for the autorenwal job to be the end date of the membership (taking into account the payment plan setting Days to renew in advance), or the minimum membership end date if there are more than one membership in the payment plan with different end dates as it used to be the case before. 

The installments receive dates  is still being calculated based on the next scheduled contribution date,

## Before

The next scheduled contribution date is what we use to trigger the autorenewal.

## After

The membership end date (or the minimum membership end date if there are more that one) is what triggers that autorenewal. Which meant reverting most of the work I did here: https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/393.

## Other notes

I also updated the migrating upgrader that updates the next scheduled contribution date for all payment plans, so it now uses the max contribution receive date plus:

- +1 year for annual payment plans
-  +1 month for monthly payment plans
-  +3 months for quarterly payment plans
-  minimum membership end date + 1 day for other manual payment plans. (I do that because having non annual, monthly or quarterly payment plans is no longer supported in v5, and it is hard to know what the next contribution date should be for these non standard payment plans, so the safest option I found is just to set it to the minimum membership end date among all the memberships attached to it, plus 1 day so equals the new membership start date if it get renewed).

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/SingleInstalmentPlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/SingleInstalmentPlan.php
@@ -20,7 +20,16 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_SingleInstalmentPlan extends C
    * - Recurring contribution has no end date.
    * - Recurring contribution is active
    * - Recurring contribution is not cancelled
-   * - next scheduled contribution date is less or equal current date (with 'days to renew in advance' setting in mind)
+   * - Recurring contribution has either:
+   *   - At least one auto-renew, un-removed line item for a membership with an
+   *     end date before today + $daysToRenewInAdvance.
+   *   - At least one auto-renew, un-removed line item and NO memberships, and
+   *     next_run_date(*) is before today + $daysToRenewInAdvance.
+   *
+   * (*) next_run_date corresponds to either maximum end date of all line items
+   * related to the recurring contribution + 1 period, or the start date of the
+   * recurring contribution + 1 period, if there are no line items with end
+   * dates.
    *
    * @return array
    */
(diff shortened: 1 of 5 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal/SingleInstalmentPlan.php
```diff
@@ -46,9 +87,12 @@ protected function getRecurringContributions() {
...
          AND msl.is_removed = 0
          AND msl.end_date IS NULL
-         AND ccr.next_sched_contribution_date IS NOT NULL
-         AND DATE(ccr.next_sched_contribution_date) <= DATE_ADD(CURDATE(), INTERVAL {$daysToRenewInAdvance} DAY)
     GROUP BY ccr.id
+      HAVING MIN(cm.end_date) <= DATE_ADD(CURDATE(), INTERVAL {$daysToRenewInAdvance} DAY)
+      OR (
+        COUNT(cm.id) = 0
+        AND next_run_date <= DATE_ADD(CURDATE(), INTERVAL {$daysToRenewInAdvance} DAY)
+      )
```
ahed-compucorp, line 95: With the `phpdoc` above made it clear to me what this query is doing, but I can see it became a bit complex and this query shares some parts with the `getRecurringContributions`'s query.

ahed-compucorp (approved):

PR #404, opened 2021-10-26
PROD-33: Add civicrm_contribution_recur to LineItem supported entity options

olayiwola-compucorp:
## Overview
CiviCRM decided to limit the supported entity_table options to some predefined tables as seen in this PR https://github.com/civicrm/civicrm-core/pull/20464, this is now causing error in this extension,  because this extension allow users to add LineItems to civicrm_contribution_recur table which is not part of the predefined tables.

## Before
343d1bc5-732a-4a6c-bdd7-fb3cee8f9b01

## After
after-line-item

## Technical Details
implement the hook_civicrm_fieldOptions hook to add civicrm_contribution_recur to the LineItem options.

### DIFF
--- a/membershipextras.php
+++ b/membershipextras.php
@@ -477,3 +477,17 @@ function _membershipextras_appendJSToModifyRecurringContributionPage(&$page) {
     ['contribution_frequency' => $frequency]
   );
 }
+
+/**
+ * Implements fieldOptions hook().
+ *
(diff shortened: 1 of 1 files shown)


omarabuhussein:
Given that this passed testing already, and that we will need it for V5, I rebased it with the current master branch (which includes V5 now) and will merge it.

PR #405, opened 2021-10-27
MAE-619: Fix error with extra amounts being added to total contribution amount

olayiwola-compucorp:
## Overview
This PR ensures that Recurring contribution amount is correct after adding a line item in manage installments view of annual payment plan and that has auto renewal

## Before
before-contrib

## After
after-contrib

## Technical Details
When calculating new total amount for recurring contribution, ensure that only amount from line items with end_dates are included (Given that the payment has one or no installment). same condition is applied to every other part of the code where total recurring contribution amount is calculated.
```php
$installments = CRM_Utils_Array::value('installments', $this->recurringContribution, 0);
if ($installments <= 1) {
    $conditions['end_date'] = ['IS NULL' => 1];
}
```


omarabuhussein (dismissed):

olayiwola-compucorp:
Merging this PR after passing QA

PR #422, opened 2022-04-27
MAE-476: Hides the auto-renew option field when the contribution tab is opened

olayiwola-compucorp:
## Overview
This PR hides the auto-renew option field when the contribution tab is opened

## Before
The auto-renew option field is visible irrespective of the current tab.
poo

## After
The auto-renew option field is only opened when the payment tab is opened. it is hidden by default.
poo-B


erawat (approved):
Nice solution using MutationObserver!

PR #423, opened 2022-05-05
MAE-684: Prevent user from user selecting membership type with different period and duration

olayiwola-compucorp:
## Overview
This PR prevents a user from selecting membership types with different `period type` and `duration unit`. should in case a user selects multiple membership types with different `period type` or `different unit` the following error is displayed `You have selected membership types with different period types (i.e. rolling and fixed) and/or different duration units. Please only select memberships with the same period types and duration units.`  

## Before
Error `All membership types must have a same period type.` is only displayed when multiple membership types with different period types are selected.

## After
The new error message is displayed when multiple membership types with different `period types` or different `duration units` are selected.
b

### DIFF
--- a/CRM/MembershipExtras/Exception/InvalidMembershipTypeInstalment.php
+++ b/CRM/MembershipExtras/Exception/InvalidMembershipTypeInstalment.php
@@ -6,5 +6,6 @@ class CRM_MembershipExtras_Exception_InvalidMembershipTypeInstalment extends Exc
   const QUARTERLY_NOT_SUPPORT = "Quarterly instalments for fixed period membership types are not supported.";
   const DURATION_INTERVAL = "Membership types must be 1 month, 1 year or 1 life time only";
   const DAY_DURATION = "Day duration unit is not supported";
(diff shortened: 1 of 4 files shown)


jamienovick:
@olayiwola-compucorp possibly this isn't the correct video for this? ;)

olayiwola-compucorp:
@olayiwola-compucorp possibly this isn't the correct video for this? ;)
yes, I selected the wrong file, it's corrected now. @jamienovick

erawat (changes requested):
Since `InvalidMembershipTypeInstalment::PERIOD_TYPE; ` is no longer in used, should we remove it from the  `InvalidMembershipTypeInstalment `class?

olayiwola-compucorp:
@olayiwola-compucorp can you also test and check that if you change the payment method and/or other items on the form the validation rule still runs? With the previous iteration I think there was a bug and this could be submitted in this case.

@jamienovick this particular validation also runs in the backend after the user submits the form.

olayiwola-compucorp:
Merging this into the feature branch, I will make a new PR to the workstream branch with updates.

PR #427, opened 2022-05-16
MAE-684: Remove validation that prevents creating membership of multiple months or years.

olayiwola-compucorp:
## Overview
In this PR, the validation that prevents users from creating rolling membership with multiple months or years duration is removed.

Also, users are prevented from creating a payment plan membership with a lifetime duration

## Before
apap2

## After
apap

Users are prevented from creating payment plan membership with life time duration
Screenshot 2022-05-16 at 14 19 01


omarabuhussein (approved):
@olayiwola-compucorp approved, but could you in the future put separate stuff into commits, give that the single commit in this PR does morre than removing the mentioned validation. Can you also update the PR title to "Improvements to membership types validation for payment plans"

PR #432, opened 2022-06-07
MAE-693:  Add support for multi years/month rolling membership

olayiwola-compucorp:
## Overview
In this PR,
- we add support for multi-year rolling membership
- we add support for multi-month rolling membership
- we ensure cycle day doesn't exceed 28
- we ensure the next scheduled contribution date day doesn't exceed 28

## Before
Multi year/month memberships can not be created
Screenshot 2022-06-07 at 10 01 28

## After
Users can now create X year membership and appropriate instalment count for the selected schedule, for example when X is 3, 36 instalments are created for the monthly schedule, 12 instalments are created for the quarterly schedule and 3 instalments are created for the annual schedule.
3year

On renewal, the membership period is extended by the membership duration, and a new recurring contribution is created.
3yearrenwal

Users can now create X month membership and appropriate instalment count for the selected schedule, for example when X is 6, 6 instalments are created for the monthly schedule.
multimonth

On renewal, the membership period is extended by the membership duration, and a new recurring contribution is created.
multimonthrenewal

Cycle day and the next scheduled contribution date day doesn't exceed 28 irrespective of the first/last contribution date
Screenshot 2022-06-08 at 12 24 52

### DIFF
--- a/CRM/MembershipExtras/Helper/InstalmentSchedule.php
+++ b/CRM/MembershipExtras/Helper/InstalmentSchedule.php
@@ -23,12 +23,16 @@ class CRM_MembershipExtras_Helper_InstalmentSchedule {
     ])['values'][0]['api.MembershipType.get']['values'][0];
 
     $durationUnit = $membershipType['duration_unit'];
-    if ($membershipType['period_type'] == 'rolling' && ($durationUnit == 'lifetime' || $durationUnit == 'month')) {
+    if ($membershipType['period_type'] == 'rolling' && ($durationUnit == 'lifetime')) {
       $instalmentDetails['instalments_count'] = 1;
+    }
+    elseif ($membershipType['period_type'] == 'rolling' && $durationUnit == 'month') {
+      $instalmentDetails['instalments_count'] = (int) $membershipType['duration_interval'];
     }
     else {
       $instalmentDetails['instalments_count'] = self::getInstalmentCountBySchedule($schedule, (int) $membershipType['duration_interval']);
     }
+
     $instalmentDetails['instalments_frequency'] = self::getFrequencyInterval($schedule);
     $instalmentDetails['instalments_frequency_unit'] = self::getFrequencyUnit($schedule, $instalmentDetails['instalments_frequency']);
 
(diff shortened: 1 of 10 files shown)


CRM/MembershipExtras/Helper/InstalmentSchedule.php
```diff
@@ -105,15 +111,15 @@ public static function isPaymentPlanPayment() {
   }
 
   public static function getPaymentPlanSchedule($frequencyUnit, $frequencyInterval, $installmentsCount) {
-    if ($frequencyUnit == 'month' && $frequencyInterval == 1 && $installmentsCount == 12) {
+    if ($frequencyUnit == 'month' && $frequencyInterval == 1) {
```
omarabuhussein, line 114: It does not seem that passing $installmentsCount is still needed

omarabuhussein (changes requested):
This looks good, so as we expected, you would able to get it work without much changes which is great, some notes though:


1- The "next scheduled contribution date" should follow the Cycle day, so in your last picture, the cycle day is 28, and the "next scheduled " date should also be on 28th of Jan.


2- I've noticed that for 3-year memberships, the list of instalments (before you click submit)
all have the same year (2019) which is obviously wrong, but given we have another ticket
to look into the list of generated instalment dates before submitting issue, I think it is fine to leave it for now until we reach that ticket, Just wanted to bring your attention to it.

omarabuhussein:
@olayiwola-compucorp I missed this during the first review, but can you also add tests to make sure the  cycle day won't exceed 28.

Also with your last commit we need  tests to ensure that "next scheduled contribution date won't exceed the 28th day"

tests/phpunit/CRM/MembershipExtras/Service/CycleDayCalculatorTest.php
```diff
@@ -0,0 +1,40 @@
...
+ * Class CRM_MembershipExtras_Service_CycleDayCalculatorTest
+ *
+ * @group headless
+ */
+class CRM_MembershipExtras_Service_CycleDayCalculatorTest extends BaseHeadlessTest {
+
+  /**
+   * Test Cycle Day is one for non monthly memberships.
+   */
+  public function testCycleDayisOneForNonMonthlyMembership() {
```
omarabuhussein, line 13: a more accurate name would be "testCycleDayDefaultsToOneForAnnualMembership"

omarabuhussein (changes requested):

PR #435, opened 2022-06-15
MAE-751: Consolidate logic for calculating membership instalment dates

olayiwola-compucorp:
## Overview
The instalment dates listed on the new membership screen can sometimes not match the contribution instalment dates after creation, this difference is noticeable when using a day above 28. In this PR we use the same logic used in calculating contribution dates to calculate the instalment dates for the new membership screen.

## Before
The instalment dates on the new membership screen.
Screenshot 2022-06-15 at 09 04 37

The new membership screen uses the inbuilt PHP DateTime function to add months to date, and this function behaves in some certain way as we've seen in the screen above, `2022-01-29 + 1month =  2022-03-01` and subsequent months use the day from the previous calculation. 

The actual instalment dates after creation.
date

Whereas when creating the contributions, the instalment dates are calculated using the custom function `getSameDayNextMonth` https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/326e69ed0468cb4f2b40ad6adc0b6630257e84d3/CRM/MembershipExtras/Service/InstalmentReceiveDateCalculator.php#L102 to add months to date, this function helps handle the scenario mentioned above such that `2022-01-29 + 1month =  2022-02-28` and subsequent dates makes use of the day from the first calculation. 

## After
New membership screen using the function `getSameDayNextMonth` to calculate instalment dates
Screenshot 2022-06-15 at 09 31 48
Screenshot 2022-06-15 at 09 32 02
Screenshot 2022-06-15 at 09 32 13

### DIFF
--- a/tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
+++ b/tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
@@ -633,7 +633,7 @@ class CRM_MembershipExtras_Service_MembershipInstalmentsScheduleTest extends Bas
   }
 
   /**
-   * Tests exception is thorwn when membership type duration is day
+   * Tests exception is thrown when membership type duration is day
    */
   public function testExceptionIsThrownIfMembershipTypeDurationUnitIsDay() {
     $invalidMembershipType = MembershipTypeFabricator::fabricate(array_merge($this->defaultRollingMembershipTypeParams,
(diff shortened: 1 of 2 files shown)


omarabuhussein (changes requested):
Can you add tests to CRM_MembershipExtras_Service_MembershipInstalmentsScheduleTest to cover these  cover such edge cases.

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
```diff
@@ -644,6 +644,30 @@ public function testExceptionIsThrownIfMembershipTypeDurationUnitIsDay() {
     $this->getMembershipInstalmentsSchedule([$membershipType], MembershipInstalmentsSchedule::ANNUAL);
   }
 
+  /**
+   * Tests that instalments date are calculated properly and months follows a sequential order.
+   */
+  public function testRollingInstalmentDatesAreCalculatedProperly() {
```
omarabuhussein, line 650: the name of the test should reflect the edge case that you are testing (which is that start date day is > 28).

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
```diff
@@ -644,6 +644,30 @@ public function testExceptionIsThrownIfMembershipTypeDurationUnitIsDay() {
...
+    );
+
+    foreach ($schedule['instalments'] as $index => $instalment) {
+      $instalmentDate = $instalment->getInstalmentDate();
+      $year = $instalmentDate->format('Y');
+      $month = $instalmentDate->format('n');
+      $day = $instalmentDate->format('d');
+
+      $this->assertTrue(checkdate($month, $day, $year));
+      $this->assertEquals($month, $currentMonth++);
```
omarabuhussein, line 667: testing the month is correct is not enough to make sure the date is calculated correctly, a more accurate test should include an array that include all the expected installment dates (all hardcoded), and you compare the resulted dates against them.

omarabuhussein (changes requested):

tests/phpunit/CRM/MembershipExtras/Service/MembershipInstalmentsScheduleTest.php
olayiwola-compucorp: Thanks, @omarabuhussein.

> a more accurate test should include an array that include all the expected installment dates (all hardcoded)

I avoided doing this, since we have tests like `testRollingMonthlyInstalmentAmounts` and others in place, and  I thought they were enough to assert the correctness of the returned date.

e.g. if the `start_date` is 2022-01-31 and the next date is 2022-03-01 the test will fail.

omarabuhussein (approved):

PR #438, opened 2022-08-03
COMPCRM-58: Return specific field from Membership API

olayiwola-compucorp:
## Overview
In this PR we return specific fields from the Membership API.

## Before
The `modify instalment` page is broken, and the user cannot perform any action.
base2

## After
The `modify instalment` page displays as expected, and the user can perform the expected actions.
base1

## Technical Details
Before now when getting related membership of line items, we return every membership field, and these might include attachment fields also, which can break the page. in this PR we limit the returned membership fields to `'start_date', 'end_date', 'id'`.

`'start_date', 'end_date', 'id'` are the membership fields we recognize are required for now.

### DIFF
--- a/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
+++ b/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php
@@ -446,6 +446,7 @@ class CRM_MembershipExtras_Page_EditContributionRecurLineItems extends CRM_Core_
       $membership = civicrm_api3('Membership', 'getsingle', [
         'sequential' => 1,
         'id' => $membershipID,
+        'return' => ['start_date', 'end_date', 'id'],
       ]);
       $membership['related_membership_type'] = $this->getMembershipTypeFromMembershipID($membershipID);
       $this->membershipsCache[$membershipID] = $membership;


omarabuhussein (changes requested):
Thanks @olayiwola-compucorp  but couple things:

1- The PR itself lacks the technical details on what is going on, I had to go to the ticket itself to get some idea why this issue happens, so can you please update the PR description with more details.

2- So it seems from what I understood, the issue happen due to us fetching the memberships for the line items in this line: https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/72f9afd871ad1af77861208d268969b4b3995121/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php#L512
which might include some custom fields that have attachments in them that ruin the screen, but I think converting the data to base64 is probably not the best approach here, in these screens we only need some of the memberships fields not all of them, so we can probably alter  the getMembership() method https://github.com/compucorp/uk.co.compucorp.membershipextras/blob/72f9afd871ad1af77861208d268969b4b3995121/CRM/MembershipExtras/Page/EditContributionRecurLineItems.php#L442 so it only returns the field we want in Current and Next period screens, which can be easily done with Civi API `return` field.

omarabuhussein (approved):

PR #452, opened 2023-02-08
CIWEMB-135: Add create payment scheme form

erawat:
## Overview

This PR adds a new create payment scheme form to the extension. 

## Before

Form did not exist. 

## After

screenshot-dev localhost_8020-2023 02 08-17_13_26

## Technical Details

The new features using new CiviCRM php8 container so when executing civix command. The civix files in the extension is updated thus, the membershipextras.php must be updated to follow the civix.php file.


omarabuhussein (commented):
@erawat , two things:

1- 
In this @olayiwola-compucorp 's PR: 

https://github.com/compucorp/uk.co.compucorp.certificate/pull/66/files

There is no `mixin` folder added after regenerating the civix file, where here they are added, I think Ola's PR is more accurate given these mixins are now part of Civi core, am I right @olayiwola-compucorp  ? 

2- I don't think the `enabled` field should be required, what happens if you create a payment scheme without it being set ?

erawat:
> @erawat , two things:
> 
> 1- In this @olayiwola-compucorp 's PR:
> 
> https://github.com/compucorp/uk.co.compucorp.certificate/pull/66/files
> 
> There is no `mixin` folder added after regenerating the civix file, where here they are added, I think Ola's PR is more accurate given these mixins are now part of Civi core, am I right @olayiwola-compucorp ?
> 
> 2- I don't think the `enabled` field should be required, what happens if you create a payment scheme without it being set ?

1. I add/update product files based on civix generated, using civix on the php8 container, @olayiwola-compucorp has tried on his local, the mixin dir is also created 
2. Updated.

omarabuhussein:
then @olayiwola-compucorp can you include the  `mixins` directory in your other PR ?  given it seems to be needed

omarabuhussein (approved):

omarabuhussein:
so as we discussed in chat, the `mixin` folder should not be included.

omarabuhussein (changes requested):
so as we discussed in chat, the mixin folder should not be included.

omarabuhussein (approved):

PR #456, opened 2023-02-27
CIWEMB-135: Add payment processor field to payment scheme form

asem-compuco:
## Overview
Add payment processor dropdown menu field where it shows only the payment processor whose type is `GoCardless` to payment scheme form

## Before
Screenshot 2023-02-27 174208

## After
Screenshot 2023-02-27 190521


erawat (changes requested):
@asem-compuco  Can we move payment processor field to before parameters field?

erawat (approved):

PR #461, opened 2023-04-06
CIWEMB-249: Add setting to allow configuring supported payment processors

omarabuhussein:
## Overview

Membershipextras only supports payment processors that implement the `Payment_Manual` class, but given we now want it to work with GoCardless as well, and in the future we will probably add Stripe support too, this approach has to change. Here we we  create new setting under  `/civicrm/admin/payment_plan_settings` called "Supported Payment Processors", that stores the payment processors that Membershipextras support.

There is another ticket to change Membershipextras codebase to use this new setting, so the work to do that is not part of this PR.

## Before

Membershipextras extension only work with Payment Processors that:

- Have their `class_name` field = `Payment_Manual`
- Or for Payment Plans with empty payment_processor_id field (which are considered "Pay Later").

## After

Nothing change regarding how Membershipextras work yet, only new setting is added under "Payment Plan settings" page,  called "Supported Payment Processors" , which stores all the payment processors that Membershipextras supports:

2023-04-06 18_07_44-Payment Plan Settings _ compuclientv2ssp

An installer and an upgrader are added as well to add "Offline Recurring Contribution" Payment processor (that is defined by Membershipextras)to the this new setting.

2023-04-06 18_08_12-Payment Plan Settings _ compuclientv2ssp

And any extension that wants to opt in to get Membershipextras support, can just add similar installer and upgrader to do so.

In future PR, this new setting will be used instead of what is mentioned in the "Before" section to determine if the payment processor is supported by Membershipextras or not.

## Other notes

-  I also updated "Offline payment processor for back office" setting field, so it now shows all the avaiable payment procssors in the system, instead of only the ones that have their `class_name` field = `Payment_Manual`.
- I also added the settings mixin to the info file, because without it the setting page will just be empty, the new Civix versions relies on the mixin to load the settings defined in the settings folder.

### DIFF
--- a/CRM/MembershipExtras/Form/PaymentPlanSettings.php
+++ b/CRM/MembershipExtras/Form/PaymentPlanSettings.php
@@ -1,5 +1,4 @@
 <?php
-use CRM_MembershipExtras_Service_ManualPaymentProcessors as ManualPaymentProcessors;
 
 /**
  * Payment Plan Settings form controller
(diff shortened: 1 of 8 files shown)


omarabuhussein:
I decided to change my approach to rely on hooks instead of settings, so I am closing this PR in favor of this one: https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/462

PR #462, opened 2023-04-07
CIWEMB-249: Add hook to allow other extensions to opt for Membershipextras support

omarabuhussein:
## Overview

Membershipextras only supports payment processors that implement the Payment_Manual class, but given we now want it to work with GoCardless as well, and in the future we will probably add Stripe support too, this approach has to change. Here we we new hook that allows such extensions to opt for Membershipextras support.

## Before

Membershipextras extension only work with Payment Processors that:

- Have their `class_name` field = `Payment_Manual`
- Or for Payment Plans with empty payment_processor_id field (which are considered "Pay Later").

## After

This extension stills supports the same conditions mentioned in the "Before" section, it also allows other payment processor extensions to add themselves to get Membershipextras support.

## Technical Details

The new hook is called:

`hook_membershipextras_updateSupportedPaymentProcessors` an passes the list of payment processors Membershipextras support  by reference `$supportedPaymentProcessor`, which allows other extension to update the list. Here is a sample implementation inside an test extension called "testextension", which adds a payment processor defined inside it called "My new Test Payment Processor" to the list of Membershipextras supported processors:

```
function testextension_membershipextras_updateSupportedPaymentProcessors(&$supportedPaymentProcessor) {
  $myExtensionPaymentProcessors = \Civi\Api4\PaymentProcessor::get()
    ->addWhere('is_test', 'IN', [FALSE, TRUE])
    ->addWhere('name', '=', 'My new Test Payment Processor')
    ->execute()
    ->getArrayCopy();

  $supportedPaymentProcessor = array_merge($supportedPaymentProcessor, $myExtensionPaymentProcessors );
}
```

## Other notes

- I also added the settings mixin to the info file, because without it the setting page will just be empty, the new Civix versions relies on the mixin to load the settings defined in the settings folder.
- Most of the changes in this PR are about changing the word "manual" (and in someplaces "offline) to "supported", given we no longer support just "Manual" or "Offline" payment processors, I don't get rid of all occurrences though, given in some places it requires more extra work than just renaming things.


CRM/MembershipExtras/Service/SupportedPaymentProcessors.php
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    return $supportedPaymentProcessorOptions;
+  }
+
+  private static function getSupportedPaymentProcessors() {
+    $supportedPaymentProcessors = self::getManualPaymentProcessors();
+
+    // Allow extensions to opt in for Membershipextras
+    // Support.
```
erawat, line 63: Not sure why we need two lines here in the inline comment.

PR #465, opened 2023-04-18
CIWEMB-217: Add new hook membershipextras_postPaymentSchemeRenewalAutoRenewal for payment scheme autorenwals

omarabuhussein:
## Before

~

## After

New hook is now  dispatched after each payment plan that is linked to payment scheme  autorenewal, the hook signature is:

```
hook_membershipextras_postPaymentSchemeRenewalAutoRenewal($paymentProcessorId, $recurContributionId, $renewalContributions)
```

where:

- $paymentProcessorId : is the payment processor id that is used to pay the payment plan
- $recurContributionId: Is the recurring contribution that got created after renewal, and that will be used for the renewed period.
- $renewalContributions: Array of contributions (instalments) that got created after the autorenewal, where each element of the array contains: `id` -> The contribution id, `total_amount` -> the contribution amount, `receive_date` -> the contribution receive date (invoice date), `currency` -> the contribution currency.

This hook will be implemented later in the GoCardless extension to manage the payments in  GoCardless after autorenewal.

### DIFF
--- a/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentSchemePlan.php
+++ b/CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentSchemePlan.php
@@ -80,6 +80,8 @@ class CRM_MembershipExtras_Job_OfflineAutoRenewal_PaymentSchemePlan extends CRM_
       $instalmentsHandler->setInstalmentsCount($this->currentRecurringContribution['installments']);
       $instalmentsHandler->createRemainingInstalments();
     }
+
+    $this->dispatchPaymentSchemeRenewalHook();
   }
 
   private function generatePaymentPlanSchedule() {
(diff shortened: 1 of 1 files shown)


CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentSchemePlan.php
```diff
@@ -327,4 +329,16 @@ protected function calculateRecurringContributionTotalAmount($recurringContribut
     return MoneyUtilities::roundToCurrencyPrecision($totalAmount);
   }
 
+  private function dispatchPaymentSchemeRenewalHook() {
+    $nullObject = CRM_Utils_Hook::$_nullObject;
+    CRM_Utils_Hook::singleton()->invoke(
+      ['paymentProcessorId', 'recurContributionId', 'paymentPlanSchedule'],
+      $this->newRecurringContribution['payment_processor_id'],
+      $this->newRecurringContributionID,
+      $this->paymentPlanSchedule,
```
erawat, line 338: @omarabuhussein  I don't think GoCardless care about payment plane schedule, what it needs is the list of contribution ID, amount, current that we want to create payments in GoCardless, which I think they were created in the part of renewal.

erawat (changes requested):

CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentSchemePlan.php
omarabuhussein: hmmm... I am not following, because that is not what was discussed and agreed on, wasn't the point of this ticket/PR https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/463 to create the API to send the data in specific format so that you can use it in GoCardless to create the payment schedule on GoCardless which also follows the format GoCardless expected ?

omarabuhussein: Update 1: I had a chat with Ererawat, and it seems that the payment schedule was needed before because we wanted not to create the contributions upfront where we create a payment schedule on GoCardless instead, but given this changed now to creating the contributions upfront, passing only the payment schedule here has no value, so I've changed the hook above to pass the contributions (instalments) list after autorenewal.

erawat (changes requested):

CRM/MembershipExtras/Job/OfflineAutoRenewal/PaymentSchemePlan.php
```diff
@@ -327,4 +329,26 @@ protected function calculateRecurringContributionTotalAmount($recurringContribut
...
+      $this->newRecurringContributionID,
+      $renewalContributions,
+      $nullObject, $nullObject, $nullObject,
+      'membershipextras_postPaymentSchemeRenewalAutoRenewal'
+    );
+  }
+
+  private function getPaymentPlanContributionsForPaymentSchemeHook() {
+    return \Civi\Api4\Contribution::get()
+      ->addSelect('id', 'receive_date', 'total_amount')
```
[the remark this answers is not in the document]
omarabuhussein: done

erawat (approved):

PR #467, opened 2023-05-18
CIWEMB-222: Only process hooks on membership payment

erawat:
## Overview

Since this https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/462, Membership Extras allows other Payment Processor to support Membership Extras via hook. 

Some payment processors like GoCardless support recurring contribution (donation subscription) that do not require membership functionality. 

This PR adds additional check if the recurring contribution links to any membership, in addition to check if processor is supported Membership Extras before executing hooks. 

## Before

Pre and post hooks always run on ContributionRecur and EntityFinancialTrxn objects, so contribution status was not recorded correctly. 

## After

Pre and post hooks for ContributionRecur and EntityFinancialTrxn objects, only run if the recurring contribution links to a membership.

### DIFF
--- a/CRM/MembershipExtras/Helper/RecurringContributionHelper.php
+++ b/CRM/MembershipExtras/Helper/RecurringContributionHelper.php
@@ -1,0 +1,21 @@
+<?php
+
+class CRM_MembershipExtras_Helper_RecurringContributionHelper {
+
+  /**
+   * Checks if recurring contribution links to any membership.
+   */
+  public static function isRecurringContributionLinkToMembership($recurringContributionID) {
+    $membershipCount = \Civi\Api4\Membership::get()
+      ->selectRowCount()
+      ->addWhere('contribution_recur_id', '=', $recurringContributionID)
+      ->execute();
+
+    if ($membershipCount->rowCount == 0) {
+      return FALSE;
+    }
+
+    return TRUE;
+  }
+
+}
--- a/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
+++ b/CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
@@ -2,6 +2,7 @@
 
 use CRM_MembershipExtras_Service_SupportedPaymentProcessors as SupportedPaymentProcessors;
 use CRM_MembershipExtras_Service_PaymentPlanStatusCalculator as PaymentPlanStatusCalculator;
+use CRM_MembershipExtras_Helper_RecurringContributionHelper as RecurringContributionHelper;
 
 /**
  * Implements pre hook on ContributionRecur entity.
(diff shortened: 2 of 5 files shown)


CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
```diff
@@ -66,18 +67,18 @@ private function getRecurringContribution($id) {
    * contribution.
    */
   public function preProcess() {
-    if ($this->operation == 'create') {
+    $isPaymentPlanPayment = CRM_MembershipExtras_Helper_InstalmentSchedule::isPaymentPlanPayment();
```
omarabuhussein, line 70: You shouldn't do this here, `CRM_MembershipExtras_Helper_InstalmentSchedule::isPaymentPlanPaymen()` checks for some form elements if they are set, where this hook is a database level hook and might run from various places at different times.

CRM/MembershipExtras/Helper/RecurringContributionHelper.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+class CRM_MembershipExtras_Helper_RecurringContributionHelper {
+
+  /**
+   * Checks if recurring contribution links to any membership.
+   */
+  public static function isRecurringContributionLinkToMembership($recurringContributionID) {
```
omarabuhussein, line 8: I would rename the method to `isRecurringContributionLinkedToMembership`

also, to correctly check if the recurring contribution is linked to a membership, you will need to do it diffeenrrly given more than one recurring contribution can be linked to a membership (specially after autorenewals), basically you will to do something close to this:

```
SELECT X FROM civicrm_contribution c
INNER JOIN civicrm_membership_payment mp ON c.id  = mp.contribution_id 
WHERE c.contribution_recur_id = {Input_recur_id}
```

omarabuhussein (changes requested):
Regarding the "cycle day" change, my initial thought was the cycle day should be calculated before dispatching the hook (which is the case before this PR), but apparently the test we have works on the assumption that it should be done after and that is also documented in this ticket: https://compucorp.atlassian.net/browse/IO-124 , so I guess I am fine with such change, though not sure how the test used to work fine before.

.github/workflows/tests.yml
```diff
@@ -49,4 +49,4 @@ jobs:
 
       - name: Run phpunit tests
         working-directory: ${{ env.CIVICRM_EXTENSIONS_DIR }}/uk.co.compucorp.membershipextras
-        run: phpunit5
+        run: phpunit9
```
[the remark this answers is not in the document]
erawat: @omarabuhussein. Ah It was a hang over before merging master branch to workstream branch. Message commit updated.

CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
erawat: @omarabuhussein  I understood that only need to process this if the payment plan was created to the UI and after our conversion to slack. I removed the code and created this ticket, so we do not forget that we would need to deal with it as it is still causing failure in GoCardless https://compucorp.atlassian.net/browse/CIWEMB-335

CRM/MembershipExtras/Helper/RecurringContributionHelper.php
erawat: @omarabuhussein  the method has been renamed, but I am not sure if I understand why we need to care about if there is any other recurring contribution links to a membership? As we just need to check if the given recurring contribution is linked to any membership?

omarabuhussein: @erawat 

it is something like this:

1- You create a membership with payment plan, at that moment the membership `contribution_recur_id` will be set to the recurring contribution id that was created.

2- After a year, the autorenewal job renewed the membership and created new recurring contribution for the new term, but in Membershipextras we don't update the membership `contribution_recur_id`  field to point to the new recurring contribution, so at the moment it basically just points to the first recurring contribution that was created. 

3- Even if we update the autorenewal job to update the membership  `contribution_recur_id` to point to the new one, it means the link between the old recurring contribution and the membership will be lost given it now will be pointing to the new one. 

because of that, using the  membership `contribution_recur_id` field is not a reliable way to tell if the recurring contribution is linked to a membership, and the only way to do it is by doing a join on the civicrm_membership_payment and the civicrm_contribution tables.

CRM/MembershipExtras/Hook/Pre/ContributionRecur.php
omarabuhussein: yeah, all I am saying here is that the approach used here is not the right one to solve the problem and might cause issues in some other places if we happen to proceed with it, so we need to come up with something else.

CRM/MembershipExtras/Helper/RecurringContributionHelper.php
omarabuhussein: @erawat on a 2nd thought, I don't think we need this new ticket, even if we let Membershipextras update GoCardless recurring contributions cycle day, it is not like CiviCRM core makes any use of it.

omarabuhussein (approved):

PR #472, opened 2023-06-06
CIWEMB-102: Allow deleting a payment scheme. 

erawat:
## Overview

This PR adds the ability to delete a payment scheme that has not been linked to any recurring contribution. 

## Before

Delete link did not work. 

## After

_Screencast shows a payment scheme cannot be deleted if it has been linked to a recurring contribution(s). The error toast message will display if the payment scheme cannot be deleted._ 

[screencast to delete payment scheme.webm](https://github.com/compucorp/uk.co.compucorp.membershipextras/assets/208713/27ac3a21-8944-44ff-a3be-97b0f3234ad9)

_Screencast shows payment scheme was deleted successfully._ 

[screencast-compucorp.atlassian.net-2023.06.06-14_31_21.webm](https://github.com/compucorp/uk.co.compucorp.membershipextras/assets/208713/27770b1e-665e-4352-b82a-57e1802c68da)

## Technical Details

In this PR, I updated the existing class `CRM_MembershipExtras_Form_PaymentScheme ` in process deletion in delete and add the logic to validate if payment scheme is linked to any recurring contribution by using ContributionRecur API v4 to check. 

```
 \Civi\Api4\ContributionRecur::get()
      ->selectRowCount()
      ->addWhere('payment_plan_extra_attributes.payment_scheme_id', '=', $this->id)
      ->execute()->count();
```

In the smarty [template](https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/472/files#diff-b98da6eab66fb5b44f3960032ec43005a95a0f03232ed1bc0e4eeebd66982af0R5), I also added the condition to display delete warning message by checking if the action equals to 8 which aligns with constant in[ Civi Action class ](https://github.com/compucorp/civicrm-core/blob/5.51.3-patches/CRM/Core/Action.php#L33)

### DIFF
--- a/CRM/MembershipExtras/Form/PaymentScheme.php
+++ b/CRM/MembershipExtras/Form/PaymentScheme.php
@@ -1,7 +1,5 @@
 <?php
 
-use CRM_MembershipExtras_ExtensionUtil as E;
-
 /**
  * Form controller class
  *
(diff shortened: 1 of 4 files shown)


CRM/MembershipExtras/Form/PaymentScheme.php
```diff
@@ -151,8 +164,54 @@ private function addPaymentProcessorField() {
...
-    $this->add('select', 'payment_processor', E::ts('Payment Processor'), $select, TRUE);
+    $this->add('select', 'payment_processor', ts('Payment Processor'), $select, TRUE);
 
   }
 
+  private function isDeleteContext() {
+    return $this->_action == CRM_Core_Action::DELETE;
+  }
+
+  private function postProcessDeletion() {
```
omarabuhussein, line 175: The logic in this method  belongs to the PaymentScheme BAO class not to the form so it would be enforced on all calls to delete a  Payment Scheme.

CRM/MembershipExtras/Form/PaymentScheme.php
```diff
@@ -151,8 +164,54 @@ private function addPaymentProcessorField() {
...
+    $contributionRecursCount = \Civi\Api4\ContributionRecur::get()
+      ->selectRowCount()
+      ->addWhere('payment_plan_extra_attributes.payment_scheme_id', '=', $this->id)
+      ->execute()->count();
+
+    if ($contributionRecursCount > 0) {
+      CRM_Core_Session::setStatus(ts('You are not able to delete this payment scheme as it’s linked with one or more recurring contributions.'), ts("Payment Scheme cannot be deleted"), "error");
```
omarabuhussein, line 182: And when I move this  method to the BAO class , I would throw an exception in the place of this, and here in the form class I would catch the exception and do the `CRM_Core_Session::setStatus` if there is an error

omarabuhussein (changes requested):

CRM/MembershipExtras/Form/PaymentScheme.php
omarabuhussein, line 175: The logic in this method belongs to the PaymentScheme BAO class not to the form so it would be enforced on all calls to delete a Payment Scheme.

CRM/MembershipExtras/Form/PaymentScheme.php
omarabuhussein, line 182: And when I move this method to the BAO class , I would throw an exception in the place of this, and here in the form class I would catch the exception and do the `CRM_Core_Session::setStatus` if there is an error

omarabuhussein (changes requested):

omarabuhussein (approved):

PR #473, opened 2023-06-07
CIWEMB-342: Better payment scheme support in the recurring contribution view page

omarabuhussein:
## Overview

Updating the recurring contribution view page, so it shows information related to payment scheme and hide unnecessary data 

## Before

On the recurring contribution page, we show the following data:

- Frequency
- Installments
- Cycle Day
- Next Scheduled Contribution Dat

appear on the recurring contribution page, even for the ones that are linked to payment schemes.

2023-06-07 22_43_25-gfsfd dfssd _ compuclient22sspv3

## After

The mentioned fields are no longer appear on the recurring contribution  view page if the recurring contribution is linked to a payment scheme, given these fields are not important for such recurring contributions, because the instalments and when they will be charged and how many of them will be created on autorenewal is no longer controlled by these parameters, and rather controlled by the payment scheme configuration.

2023-06-07 22_45_24-gfsfd dfssd _ compuclient22sspv3

also in the section marked in red above, a new table field is added, that shows the projected instalments after autorenewal and their expected charge dates, but it is only shown if the recurring contribution is linked to a payment scheme.


js/modifyRecurringContributionPage.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    });
+  }
+
+  function addFuturePaymentSchemeSchedule() {
+    if (payment_scheme_schedule === null) {
+      return;
+    }
+
+    var payment_scheme_block = '<tr><td class="label">Future payment scheme Instalments</td><td>';
```
erawat, line 31: Don't we localise these texts? 

Perhaps, it is better if we put the table in the smarty template, so it is faster than concatenating string in browser the content on run time?

js/modifyRecurringContributionPage.js
```diff
@@ -0,0 +1,50 @@
+CRM.$(function () {
+  const payment_scheme_schedule = CRM.vars.membershipextras.payment_scheme_schedule;
+  hideUnnecessaryPaymentPlanFields();
+  addFuturePaymentSchemeSchedule();
+
+  function hideUnnecessaryPaymentPlanFields() {
+    const frequency = CRM.vars.membershipextras.recur_contribution['frequency_unit'];
+    const payment_scheme_fields_to_hide = ['Cycle Day', 'Frequency', 'Installments', 'Next Contribution'];
```
erawat, line 8: @omarabuhussein  I understand that we do not have a variable name convention in JavaScript/jQuery for our extension development.  In other field in the ME extension, the camel case has been used in this function. Perhaps, we should use a camel case to keep variable consistent with other files?

erawat (changes requested):

js/modifyRecurringContributionPage.js
omarabuhussein: > Don't we localise these texts?

While this is true, for all the time I've been here at Compuco we never had to do translation for any client nor any one ever created a github issue asking for translation support or anything, so I am happy to skip it.

> Perhaps, it is better if we put the table in the smarty template, so it is faster than concatenating string in browser the content on run time?

I really don't see that there will be  much of performance improvement (if any) to justify changing the approach.

js/modifyRecurringContributionPage.js
omarabuhussein: fixed

erawat (approved):

jamienovick:
looks great @omarabuhussein !! ❤️

PR #476, opened 2023-06-19
CIWEMB-269: Simplify switching payment plan membership from one type to another

omarabuhussein:


omarabuhussein:
@jamienovick can you check the stuff above if you are happy with how things look like, specially the financial entries. 

I also have some questions:

1- Here and as written in the PR description above, I adjust the instalments (contributions) with date >= the selected Switch Date which I think make sense, an alternative approach would be to adjust the future ones only with no respect to the switch date, so any contribution with date >= Today Date will get adjusted, I think using the switch date is what the user will expected, what do you think ?

2- When I create the new financial entries, the "Transaction Date" = "Today Date" (the date in which the user did the membership type change) given it is the actual date of the transaction, but do you think it make more sense for it to be = the contribution date ? what is more accurate form an accounting perceptive ?

jamienovick:
@omarabuhussein, great write up. In answer to your questions:

> 1- Here and as written in the PR description above, I adjust the instalments (contributions) with date >= the selected Switch Date which I think make sense, an alternative approach would be to adjust the future ones only with no respect to the switch date, so any contribution with date >= Today Date will get adjusted, I think using the switch date is what the user will expected, what do you think ?

Correct, I agree that using the switch date would be more correct. If there is a switch date in the past however we can use all the contributions after that date (for this recurring contribution). If some of those contributions are completed already I would ignore them (if thats possible? - see my point 5 also).

> 2- When I create the new financial entries, the "Transaction Date" = "Today Date" (the date in which the user did the membership type change) given it is the actual date of the transaction, but do you think it make more sense for it to be = the contribution date ? what is more accurate form an accounting perceptive ?

I think we should actually use the contribution date. The amount isn't actually due yet so it would be more correct for the accounting "revenue recognition" to take place on the invoice date which is the contribution date.

A few other things:

3) Period end date as min vs max of membership end dates

> the "End Date" will be the period end date. The "Period end date" here means the Minimum membership end date across all the membership line items in the payment plan

so I actually think this should be the maximum end date. I appreciate what you said about "Though and ideally, all memberships within the payment plan should have the same duration and the same "end date", and we introduced some validations long ago on that, but it is not always easy to guarantee that given there are many ways in which membership end dates can be changed in CiviCRM." 

I think we should ignore for now the fact that people may edit the memberships incorrectly, but I think it's always been intended that for the view/modify future instalments to work, the end date (for renewing memberships) needs to be the same. 

nb. Perhaps as a separate ticket we can explore the logic again about the autorenew job and whether renewing based on the minimum end date is the best approach but lets leave that out of this ticket for now.

4) Should we validate when someone selects a start date in the future whether the "future start" membership status rule is enabled? If not I think you might get an error d

5) Relating to point 2, should we validate that there are suitable contributions available to add the fee too, and if not we shouldn't show this option? (maybe this is in another ticket in the spec somewhere?)

omarabuhussein:
@jamienovick 

> Correct, I agree that using the switch date would be more correct. If there is a switch date in the past however we can use all the contributions after that date (for this recurring contribution). If some of those contributions are completed already I would ignore them.  (if thats possible?)  see my point 5 also).

Yeah I already ignore any contribution that is not "Pending", I updated the PR description to make that more clear.

> Relating to point 2, should we validate that there are suitable contributions available to add the fee too, and if not we shouldn't show this option? (maybe this is in another ticket in the spec somewhere?)

At the moment it will throw an error that says "No pending instalments found for the payment plan after the selected date: X-Y-Z" which I think is enough, right ? 


> I think we should actually use the contribution date. The amount isn't actually due yet so it would be more correct for the accounting "revenue recognition" to take place on the invoice date which is the contribution date.

Make sense, I will adjust the code according to this.


> I think we should ignore for now the fact that people may edit the memberships incorrectly, but I think it's always been intended that for the view/modify future instalments to work, the end date (for renewing memberships) needs to be the same.
> nb. Perhaps as a separate ticket we can explore the logic again about the autorenew job and whether renewing based on the minimum end date is the best approach but lets leave that out of this ticket for now.

Okay then, will leave things as is for now.

> 
Should we validate when someone selects a start date in the future whether the "future start" membership status rule is enabled? If not I think you might get an error.

We can, but maybe instead we can make "Future Start" and "Current Renewed" status rules both "Reserved" , so they are always there, or is that too aggressive ?

jamienovick:
Trying to continue with the numbering of the points below:

1) Switch date and 1.1) ignore pending contributions when splitting the amount over future instalments

> Correct, I agree that using the switch date would be more correct. If there is a switch date in the past however we can use all the contributions after that date (for this recurring contribution). If some of those contributions are completed already I would ignore them. (if thats possible?) see my point 5 also).

> Yeah I already ignore any contribution that is not "Pending", I updated the PR description to make that more clear.

Great thanks

2) Use contribution date for transaction date

> I think we should actually use the contribution date. The amount isn't actually due yet so it would be more correct for the accounting "revenue recognition" to take place on the invoice date which is the contribution date.

> Make sense, I will adjust the code according to this.

Great thanks

3) Period end date as min vs max of membership end dates

I'm not sure if you also saw my point that we should use the max date and not the min date?

4) Validate we have future start date

> Should we validate when someone selects a start date in the future whether the "future start" membership status rule is enabled? If not I think you might get an error.

> We can, but maybe instead we can make "Future Start" and "Current Renewed" status rules both "Reserved" , so they are always there, or is that too aggressive ?

Hmmm, I guess I'm not close enough to how these really work, what the knock on effects might be and so I'd be uncertain I'd want to do that as part of this ticket. I think to be safe maybe lets add the validation instead?

5) Validate contributions are available

> Relating to point 2, should we validate that there are suitable contributions available to add the fee too, and if not we shouldn't show this option? (maybe this is in another ticket in the spec somewhere?)

> At the moment it will throw an error that says "No pending instalments found for the payment plan after the selected date: X-Y-Z" which I think is enough, right ?

Sounds perfect.

---

Did I get everything? Please see point 3 however, think that is important.

omarabuhussein:
> 3. I'm not sure if you also saw my point that we should use the max date and not the min date?

I did but I did not think it matter given you also wrote "I think we should ignore for now the fact that people may edit the memberships incorrectly" and that as long as the autorenewal works the way it work at the moment, it won't make a difference. but fine I guess I can change it to use the Max end date instead. 

> 4. Hmmm, I guess I'm not close enough to how these really work, what the knock on effects might be and so I'd be uncertain I'd want to do that as part of this ticket. I think to be safe maybe lets add the validation instead?

fine I guess, but that I think of it again, I am not so keen on adding such validation, the "future start" status rule is not something special, anyone can create a new status rule in CiviCRM with similar configuration to get similar effect, or even its name could be changed, so adding a validation that checks for its exitance does not seem right? you disagree I guess ?

jamienovick:
3. Max date

Yes, sorry, I see that saying "I think we should ignore for now the fact that people may edit the memberships incorrectly" could be unclear. 

3.1 I think it's essential we use the max date (as this is what we are suggesting is the end of the period). 

3.2 I think we could in future look to do some better validation of stopping people editing the end dates of memberships which are in payment plans so that the end dates cannot be different and have multiple memberships autorenewing, but I think we can improve MAE at a future time to do so.

4. Future start

Agreed, but I don't know what the impact might be on existing sites by enabling it. I'm not sure if I've road tested it enough. If you feel we shouldn't have the validation I'm happy to live without it. People will just get an error I guess anyway when they try and switch the membership and use a date in the future if it's not enabled?

omarabuhussein:
> 4. Future start: People will just get an error I guess anyway when they try and switch the membership and use a date in the future if it's not enabled?

Exactly, an error will appear telling the user that the system cannot figure out the status for the selected dates, similar to the one you get on a standard CiviCRM site with the standard Civi core membership status rules.

jamienovick:
> 4. Future start: People will just get an error I guess anyway when they try and switch the membership and use a date in the future if it's not enabled?
> Exactly, an error will appear telling the user that the system cannot figure out the status for the selected dates, similar to the one you get on a standard CiviCRM site with the default membership status rules.

Sounds just like the validation rule I was hoping for! :)

omarabuhussein:
both 

> 2. Use contribution date for transaction date

and

>  3. Max date


Are now done, and I've updated the PR description.

erawat (approved):

PR #482, opened 2023-07-19
: Fix payment scheme listing table

ahed-compucorp:
## Overview
This PR updates the payment scheme listing table to :
- Re-order the columns to match the spec. file.
- Display the permission label i.e. capitalize the first letter.
- Remove the uppercase effect from the table header.
- Add the ID column.
- Update the table styles to match the Shoreditch theme.

## Before
Screenshot from 2023-07-19 12-52-45

## After
Screenshot from 2023-07-19 13-53-47


jamienovick:
@ahed-compucorp 
Looks great, but a question, where did we add the styles? Should they be in Shoreditch itself perhaps?

Also I noticed the buttons are in the table. Should they be separate as they are probably not part of that component?

Thanks

ahed-compucorp:
@jamienovick 
After looking at Shoreditch theme, the styles there are specific to certain [pages](https://github.com/compucorp/org.civicrm.shoreditch/blob/85a77a34f1777f5cc6a4e82791d4e4dbed5503ae/scss/civicrm/administer/permissions/_access.scss#L3) and [tables](https://github.com/compucorp/org.civicrm.shoreditch/blob/85a77a34f1777f5cc6a4e82791d4e4dbed5503ae/scss/civicrm/administer/contribute/_contribute.scss#L84). The general styles that got applied to this table (or any new table created by an extension)  doesn't account for the buttons, even the width of each column didn't match the similar table like the one in "Manage ACL Roles" :

Screenshot from 2023-07-19 15-47-44

> Should they be in Shoreditch (Bootstrap) itself perhaps so that all future search kits get those styles?
Extensions could benefit from updating the "general styles". I'll work on that.

> Should they be separate as they are probably not part of that component?
I've done that after looking at other tables like the above one "Manage ACL Roles" and below another table example:

Screenshot from 2023-07-19 15-45-59

Some tables have a button at the bottom of the table.

jamienovick:
@ahed-compucorp 
Do you think we need to have a call to discuss? 

I really don't think we should be creating styles for one individual page, especially those that are bootstrap (this is BS?) and using search kit.

I think the example screens you are using are not search kit screen outputs. For example this screen in the latest CiviCRM is created by search kit and is the sort of thing we are targeting to work the same?

https://dmaster.demo.civicrm.org/civicrm/admin/custom/group?reset=1

Built with:
This search kit:
https://dmaster.demo.civicrm.org/civicrm/admin/search#/edit/4
and
Then embedded in this afform:
https://dmaster.demo.civicrm.org/civicrm/admin/afform#/edit/afsearchAdminCustomFields

If the concern is just around the button placement, I think we should just have those at the top separate as a separate component. They don't need to be styled as part of the table. 2 buttons at the top above the table would be fine.

Does that make sense or need a call?

ahed-compucorp:
@jamienovick 
> Do you think we need to have a call to discuss?

Let's have a call.

> I think the example screens you are using are not search kit screen outputs

Yes, you're right. CiviCRM migrated some of the pages/tables to search kit, I wasn't aware of that.

> I really don't think we should be creating styles for one individual page, especially those that are bootstrap (this is BS?) and using search kit.

Got it, I'll try using the search kit to render this table.

> Shoreditch has 2 themes....

Thank for the info.

ahed-compucorp:
> Got it, I'll try using the search kit to render this table.

I looked into this, but I think it is not clear how to featurize the search-kit-related configurations.

erawat (approved):

PR #490, opened 2023-08-07
CIWEMB-433: Fix extending add-on memberships when using payment.create API

omarabuhussein:
## Overview

This is a follow up to this PR  https://github.com/compucorp/uk.co.compucorp.membershipextras/pull/485 where I fixed an issue where payment plan memberships gets extended when paying for any pending instalment using the Payment.create API.

We discovered that if there are more than one membership linked to the same instalment (like if a price set was used), then and despite my work on the PR above will result in extending other memberships except the first one when that instalment gets paid for using the Payment.create API.

Reason for this problem, is that I relied on a session variable `Civi::$statics[ExtensionUti::LONG_NAME]['paymentApiCalled']` to determine if the membership is being edited from Payment.create API context, and if that is the case then I right away unset this variable, I did that as I thought to myself that it is no longer needed after I determine if the membership is being edited from Payment API context , so no need to keep it stored in session, but apparently and in case of multiple memberships, the Membership pre Edit hook will be called for every membership that is linked to the instalment, so my fix here was just to remove this line that unsets the session variable above.


ahed-compucorp (approved):
I don't think it is right to call the Civi::$statics[ExtensionUti::LONG_NAME]['paymentApiCalled'] a 
session variable, because these values will be lost in the subsequent request.

omarabuhussein:
ah you are right actually, it is really just a global variable.

PR #517, opened 2024-01-31
MAE-1114: Fix The Extension Installation on some sites

shahrukh-compuco:
## Overview
This PR fixes the extension installation on some sites.

## Before
Previously the extension installation was failing if the **EFT** payment method was not there as the creation of manual payment processor was dependent on **EFT** payment method.

## After

This PR changes the flow to first create a new payment by the name of **Offline Automated Recurring** method that manual payment processor will use instead of **EFT**

## Technical Overview
We create a new payment method attached to default financial account of type **Asset** or the first financial account of type **Asset** if default financial account of type **Asset** is not present

### DIFF
--- a/CRM/MembershipExtras/Setup/Manage/ManualPaymentProcessor.php
+++ b/CRM/MembershipExtras/Setup/Manage/ManualPaymentProcessor.php
@@ -24,7 +24,7 @@ class CRM_MembershipExtras_Setup_Manage_ManualPaymentProcessor extends AbstractM
       'is_default' => '0',
       'class_name' => 'Payment_Manual',
       'is_recur' => '1',
-      'payment_instrument_id' => 'EFT',
+      'payment_instrument_id' => (new CRM_MembershipExtras_Setup_Manage_OfflineRecurringPaymentMethod)->getPaymentMethod(),
     ];
 
     // creates the live version of the payment processor.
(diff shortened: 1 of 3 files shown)


CRM/MembershipExtras/Service/PaymentMethod.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
+use Civi\Api4\EntityFinancialAccount;
+
+class CRM_MembershipExtras_Service_PaymentMethod {
```
omarabuhussein, line 5: This is not a service, it just a class that creates a payment method, you should instead move it to this folder:
https://github.com/compucorp/uk.co.compucorp.membershipextras/tree/a09d5ea648c9c7e1bec98f7910343dc4dc541a13/CRM/MembershipExtras/Setup/Manage

also probably you should rename it to:

`CRM_MembershipExtras_Setup_Manage_OfflineRecurringPaymenMethod` instead of `CRM_MembershipExtras_Setup_Manage_PaymenMethod`
so it is more clear what this calss does, once you do that you can actually reduce this constant name from:
MANUAL_PAYMENT_PROCESSOR_PAYMENT_METHOD_NAME

to just

NAME

omarabuhussein (changes requested):

omarabuhussein:
hence that the tests are failing for a reason that is unrelated to the work in this PR, so if it passes testing we can merge it, I will need to look into these failed tests separately

omarabuhussein (approved):

PR #549, opened 2025-03-11
CIWESPRT-47: Fix Financial Item Creation For Payment Plan Contributions

shahrukh-compuco:
## Overview
Currently when creating a membership record with payment plan or duplicating an existing contribution the tax financial item does not get created, this PR fixes this issue.

## Before
Financial item for tax was missing for payment plan contributions except for the first contribution.

## After
Tax financial item gets created for every contribution in a payment plan

## Technical Details
There were checks in the code for checking if line item have tax and contribution record also have tax then only create the financial iem for tax. This check was wrong and we should only check if the line item has the tax then we should create the tax financial item because the contribution record does not have tax anymore I think it might have been something that was being done in past(that's an assumption as I could not find any related change) even normal one off tax eligible contributions also does not have anything in tax_amount column in contribution table and even [core code](https://github.com/civicrm/civicrm-core/blob/master/CRM/Price/BAO/LineItem.php#L416) only checks for tax amount in line item for creating tax financial item.


erawat (changes requested):
@shahrukh-compuco  

Can you explain more about what you mentioned—did you do it from the UI? I see that the classes you changed are "Duplicate" and "Offline Auto-Renewal Payment Plan."

Can we add examples of transaction records before and after, as well as UI + the unit tests that cover both scenarios?

erawat (dismissed):

erawat (approved):

PR #578, opened 2026-01-29
COMCL-1345: Fix Price Set Amount Display

shahrukh-compuco:
## Overview
This PR fixes the display of price set amount on new membership back office form which currently displays NAN when the amount is in thousands.

## Before
screen_recording_before

## After
screen_recording_after

## Technical Details
This issue was caused by due to the fact that we were formatting the total amount twice, once in backend then in frontend as well which caused NAN as the total amount was already formatted and contained a comma in it. This PR fixes the issue by using the correct format method only in backend.


erawat (approved):
Good fix — the double-formatting diagnosis is correct and the solution is clean. Just verify that `CRM_Utils_Money::format(..., NULL, NULL, TRUE)` returns a locale-aware formatted string (with commas) and that all other callers of `updateTotalAmount` with `isPriceSet=true` now pass a pre-formatted value since `CRM.formatMoney()` is no longer applied in JS.

PR #592, opened 2026-08-13
CIVIPLMMSR-731: Block webform end-date extension of memberships with unpaid renewal payment

erawat:
## Overview
Safety net for CIVIPLMMSR-731: online membership renewals submitted through a webform with an offsite payment processor (Stripe, GoCardless) extend the membership end date before any payment completes, and every duplicate submission adds another unpaid term. This change blocks the end-date extension at the `hook_civicrm_pre` level while the membership's renewal payment is incomplete, protecting all sites immediately while the webform_civicrm fix rides the release cycle.

## Before
A webform renewal submission against an unpaid contribution moves the membership end date forward one term per submission (confirmed 6 years from one £43 payment on woodcraft — see CIVIPLMMSR-731). The existing `preventExtendingPaymentPlanMembership()` guard covers exactly this scenario, but only for payment-plan memberships: `isSupportedPaymentPlanMembership()` returns FALSE when there is no recurring contribution, so one-off webform renewals are never protected.

## After
On a Membership edit, the end-date change is stripped (same mechanism as the payment-plan guard) when **all** of the following hold:

* the request is a Drupal webform submission (`form_id` = `webform_client_form_*` / `webform_submission*`);
* it is not a payment-receiving context (Payment API, record payment, or core's payment-completion path via the `membership_activity_status = Completed` marker);
* the membership is **not** on a payment plan (those keep the existing guard);
* the parameters move the end date forward;
* the membership's latest linked contribution is not Completed.

Legitimate flows are unaffected: payment completions via IPN/webhook arrive in separate requests with no webform `form_id`; same-request completions are exempted by the payment-context checks; staff edits from CiviCRM admin forms are not webform submissions.

## Technical Details
* `CRM/MembershipExtras/Hook/Pre/MembershipEdit.php`: new `isUnpaidWebformRenewal()` decision (called from `preProcess()` when the payment-plan guard did not already fire) plus `preventExtendingUnpaidMembership()`, which reuses the established strip mechanism (`unset end_date` + suppress the renewal activity).
* "Unpaid" is determined from the latest contribution linked via `civicrm_membership_payment`. Note core's `completeOrder` updates the membership **before** saving the contribution as Completed, which is why completion contexts must be exempted explicitly rather than relying on the stored contribution status.
* Four headless tests cover: blocked when the latest linked payment is Pending; allowed when Completed; allowed outside webform requests; allowed in the payment-completion context.

### Core overrides
None.

## Comments
* This is a mitigation/backstop, not the root-cause fix: it cannot block the very first unpaid grant (no linked contribution exists yet at that point). The root-cause fix is in webform_civicrm (https://github.com/compucorp/webform_civicrm/pull/96); together they make the whole class of bug unreachable.
* Full analysis and production evidence on [CIVIPLMMSR-731](https://compucorp.atlassian.net/browse/CIVIPLMMSR-731).

### DIFF
--- a/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
+++ b/CRM/MembershipExtras/Hook/Pre/MembershipEdit.php
@@ -89,6 +89,10 @@ class CRM_MembershipExtras_Hook_Pre_MembershipEdit {
       $this->preventExtendingPaymentPlanMembership();
     }
 
+    if (!$preventMembershipDateExtension && $this->isUnpaidWebformRenewal()) {
+      $this->preventExtendingUnpaidMembership();
+    }
+
     if (!in_array($this->id, self::$extendedMemberships)) {
       $isPaymentPlan = $this->isPaymentPlanBeingRecordedOnForm();
       $isMembershipRenewal = CRM_Utils_Request::retrieve('action', 'String') & CRM_Core_Action::RENEW;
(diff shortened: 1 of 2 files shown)


erawat:
Closing. The root-cause fix is in webform_civicrm (compucorp/webform_civicrm#96): an unpaid one-off renewal no longer changes the membership at submission, and core grants a single term on payment completion. With that in place this guard is inert on the normal path, so no change to membershipextras is needed.

